PR #2, opened 2024-02-29
fix: do not coerce `undefined` to `null` for JSON serialization

achingbrain:
Removes the CID decoder and disallows `undefined` when decoding CBOR as these do not round-trip to JSON.

Expands the test suite to be explicit about the types that are not supported.

## Change checklist


2color (approved):
LGTM!

PR #3, opened 2024-02-29
fix: append query path to path resolved from IPNS name

achingbrain:
If an IPNS name resolves to a CID+path, and the requested URL also has a path, the resolved path should be the path from the IPNS record joined with the path from the requested URL, not the other way round.

- Reduces test duplication
- Moves the `parse-url-string.spec.ts` test to `test/utils` so it mirrors the `src/utils` location of the `parse-url-string.ts` file it's testing, this way it's easy to find what is testing what

## Change checklist


packages/verified-fetch/test/utils/parse-url-string.spec.ts
```diff
@@ -309,15 +270,42 @@ describe('parseUrlString', () => {
...
       expect(result.protocol).to.equal('ipns')
       expect(result.cid.toString()).to.equal('QmQJ8fxavY54CUsxMSx9aE9Rdcmvhx8awJK2jzJp4iAqCr')
       expect(result.path).to.equal('some/path/to/file.txt')
       expect(result.query).to.deep.equal({ fomat: 'dag-cbor' })
     })
+
+    it('should parse an ipns:// url with a path that resolves to a CID with a path', async () => {
+      const cid = CID.parse('QmQJ8fxavY54CUsxMSx9aE9Rdcmvhx8awJK2jzJp4iAqCr')
+      const peerId = await createEd25519PeerId()
+      const recordPath = 'foo'
```
2color, line 291: Do you think it might make sense to add another test with:

      const recordPath = 'foo/'

To ensure there's no funky business with the slashes?

achingbrain: Good shout, I've added some more tests

2color: Nice!

SgtPooki (commented):
Nice

PR #4, opened 2024-02-29
feat: support IPFS/IPNS paths, Gateways, etc

achingbrain:
Adds support for fetching resources in the form:

- IPFS Path, e.g. `/ipfs/<CID>`
- IPNS Path, e.g. `/ipns/<PeerId>`
- IPFS Gateway Path, e.g. `http://example.com/ipfs/<CID>`
- IPNS Gateway Path, e.g. `http://example.com/ipns/<PeerId>`
- IPFS Subdomain Gateway Path, e.g. `http://<CID>.ipfs.example.com`
- IPNS Subdomain Gateway Path, e.g. `http://<PeerId>.ipns.example.com`

Notes:

- Supports paths & queries as per `ipfs://` and `ipns://` URLs.
- Uses all available methods to resolve a given CID to some content - it does not, for example, when given a subdomain gateway path, restrict downloading the blocks from only that subdomain gateway

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/utils/parse-url-string.ts
+++ b/packages/verified-fetch/src/utils/parse-url-string.ts
@@ -33,7 +33,7 @@ export interface ParsedUrlStringResults {
 const URL_REGEX = /^(?<protocol>ip[fn]s):\/\/(?<cidOrPeerIdOrDnsLink>[^/?]+)\/?(?<path>[^?]*)\??(?<queryString>.*)$/
 const PATH_REGEX = /^\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/?]+)\/?(?<path>[^?]*)\??(?<queryString>.*)$/
 const PATH_GATEWAY_REGEX = /^https?:\/\/(.*[^/])\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/?]+)\/?(?<path>[^?]*)\??(?<queryString>.*)$/
-const SUBDOMAIN_GATEWAY_REGEX = /^https?:\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\.(?<protocol>ip[fn]s)\.([^/?]*)\/?(?<path>[^?]*)\??(?<queryString>.*)$/
+const SUBDOMAIN_GATEWAY_REGEX = /^https?:\/\/(?<cidOrPeerIdOrDnsLink>[^/.?]+)\.(?<protocol>ip[fn]s)\.([^/?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
 
 function matchURLString (urlString: string): Record<string, string> {
   for (const pattern of [URL_REGEX, PATH_REGEX, PATH_GATEWAY_REGEX, SUBDOMAIN_GATEWAY_REGEX]) {
@@ -44,7 +44,7 @@ function matchURLString (urlString: string): Record<string, string> {
     }
   }
 
-  throw new TypeError(`Invalid URL: ${urlString}, please use ipfs://, ipns://, or gateway URLs only.`)
+  throw new TypeError(`Invalid URL: ${urlString}, please use ipfs://, ipns://, or gateway URLs only`)
 }
 
 /**
--- a/packages/verified-fetch/src/utils/parse-url-string.ts
+++ b/packages/verified-fetch/src/utils/parse-url-string.ts
@@ -31,6 +31,21 @@ export interface ParsedUrlStringResults {
 }
 
 const URL_REGEX = /^(?<protocol>ip[fn]s):\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
+const PATH_REGEX = /^\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
+const PATH_GATEWAY_REGEX = /^http(s?):\/\/(.*)\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
+const SUBDOMAIN_GATEWAY_REGEX = /^http(s?):\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\.(?<protocol>ip[fn]s)\.([^/?]*)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
+
+function matchURLString (urlString: string): Record<string, string> {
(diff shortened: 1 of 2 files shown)


achingbrain:
After https://github.com/ipfs/helia-verified-fetch/pull/5 we will have the `.url` property on responses that is the canonical URL for the requested resource.

With this PR the user can request the same resource via multiple URL formats.  Should we be rewriting this URL to turn, for example, subdomain gateway requests into `ipfs://` URIs?

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -31,6 +31,21 @@ export interface ParsedUrlStringResults {
 }
 
 const URL_REGEX = /^(?<protocol>ip[fn]s):\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
+const PATH_REGEX = /^\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
```
2color, line 34: Why are the `$` necessary in the negated sets of the named capture groups, i.e., `[^/$?]` are we expecting $ in the URL?

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -31,6 +31,21 @@ export interface ParsedUrlStringResults {
...
+const PATH_GATEWAY_REGEX = /^http(s?):\/\/(.*)\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
```
2color, line 35:
```suggestion
const PATH_GATEWAY_REGEX = /^http[s?]:\/\/(.*[^\/])\/(?<protocol>ip[fn]s)\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
```

You probably don't need to capture the s in http and avoid matching everything in the hostname part of the URL.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -31,6 +31,21 @@ export interface ParsedUrlStringResults {
...
+const SUBDOMAIN_GATEWAY_REGEX = /^http(s?):\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\.(?<protocol>ip[fn]s)\.([^/?]*)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
```
2color, line 36:
```suggestion
const SUBDOMAIN_GATEWAY_REGEX = /^http(s?):\/\/(?<cidOrPeerIdOrDnsLink>[^/$\.?]+)\.(?<protocol>ip[fn]s)\.([^/?]*)\/?(?<path>[^$?]*)\??(?<queryString>.*)$/
```

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -31,6 +31,21 @@ export interface ParsedUrlStringResults {
...
+function matchURLString (urlString: string): Record<string, string> {
+  for (const pattern of [URL_REGEX, PATH_REGEX, PATH_GATEWAY_REGEX, SUBDOMAIN_GATEWAY_REGEX]) {
+    const match = urlString.match(pattern)
+
+    if (match?.groups != null) {
+      return match.groups
+    }
+  }
+
+  throw new TypeError(`Invalid URL: ${urlString}, please use ipfs:// or ipns:// URLs only.`)
```
2color, line 47:
```suggestion
  throw new TypeError(`Invalid URL: ${urlString}, please use ipfs://, ipns://, or gateway URLs only.`)
```

packages/verified-fetch/src/utils/parse-url-string.ts
achingbrain: I don't think the `[s?]` is necessary either, just `?` seems to work.

packages/verified-fetch/src/utils/parse-url-string.ts
achingbrain: I'm not sure why they are there, it doesn't look like they are necessary and the tests pass without them, have removed.

packages/verified-fetch/src/utils/parse-url-string.ts
achingbrain: I've made this change, though without the `$`

SgtPooki:
> With this PR the user can request the same resource via multiple URL formats. Should we be rewriting this URL to turn, for example, subdomain gateway requests into ipfs:// URIs?

I think we would want to just parse the information we need.. 

---

some comments unrelated to the above, but more for the change this PR brings in general:

I was hoping this library could stick with `ipfs://` and `ipns://` as strictly as possible so we could more officially advocate for these protocol urls until we got native browser support, but the DX argument for supporting https & http urls is pretty strong. This change will make using this library much simpler for consumers.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -41,13 +56,7 @@ const URL_REGEX = /^(?<protocol>ip[fn]s):\/\/(?<cidOrPeerIdOrDnsLink>[^/$?]+)\/?
...
 export async function parseUrlString ({ urlString, ipns, logger }: ParseUrlStringInput, options?: ParseUrlStringOptions): Promise<ParsedUrlStringResults> {
   const log = logger.forComponent('helia:verified-fetch:parse-url-string')
-  const match = urlString.match(URL_REGEX)
-
-  if (match == null || match.groups == null) {
-    throw new TypeError(`Invalid URL: ${urlString}, please use ipfs:// or ipns:// URLs only.`)
-  }
-
-  const { protocol, cidOrPeerIdOrDnsLink, path: urlPath, queryString } = match.groups
+  const { protocol, cidOrPeerIdOrDnsLink, path: urlPath, queryString } = matchURLString(urlString)
```
SgtPooki, line 59: nice improvement here

packages/verified-fetch/test/utils/parse-url-string.spec.ts
```diff
@@ -357,15 +566,275 @@ describe('parseUrlString', () => {
...
+            query: {
+              format: 'tar'
+            }
+          }
+        )
+      })
+    })
+  })
+
+  HTTP_PROTOCOLS.forEach(proto => {
```
SgtPooki, line 754: I was `CTRL+F`ing for `http` and `https` and it took me way too long to find this.

packages/verified-fetch/test/utils/parse-url-string.spec.ts
```diff
@@ -357,15 +566,275 @@ describe('parseUrlString', () => {
...
+    describe(`${proto}://<CID>.ipns.example.com URLs`, () => {
```
SgtPooki, line 755:
```suggestion
    describe(`${proto}://<key>.ipns.example.com URLs`, () => {
```

SgtPooki (approved):
minor nit on test group name, I like that it didn't take much code (besides a ton of tests) to add this support though!

PR #5, opened 2024-02-29
feat: support redirects for UnixFS directories

achingbrain:
Adds support for simulating redirects for UnixFS directories.

We're somewhat in uncharted water here because `window.fetch` does this transparently unless you specify a [redirect option](https://developer.mozilla.org/en-US/docs/Web/API/fetch#redirect), none of which actually allow you to manually follow a redirect.

The states we can be in are:

1. URL: `ipfs://QmFoo/dir/`
  - Happy path
  - 200 response
  - `response.redirected = false`
  - `response.url = 'ipfs://QmFoo/dir'`

2: URL: `ipfs://QmFoo/dir`, `redirect: 'follow'`
  - The default value
  - Simulates automatically following a redirect
  - 200 response
  - `response.redirected = true`
  - `response.url = 'ipfs://QmFoo/dir/'`

3: URL: `ipfs://QmFoo/dir`, `redirect: 'error'`
  - Throw an error if a redirect would take place
  - Throws `TypeError('Failed to Fetch')` same as `window.fetch`

4: URL: `ipfs://QmFoo/dir`, `redirect: 'manual'`
  - Allows a caller to take action on the redirect
  - 301 response
  - `response.redirected = false`
  - `response.url = 'ipfs://QmFoo/dir`
  - `response.headers.get('location') = 'ipfs://QmFoo/dir/'`

Number 4 is the furthest from [the fetch spec](https://fetch.spec.whatwg.org/#concept-request-redirect-mode) but to follow the spec would make it impossible to actually follow a redirect.

Refs: https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62

## Change checklist

### DIFF
--- a/packages/verified-fetch/README.md
+++ b/packages/verified-fetch/README.md
@@ -408,6 +408,79 @@ console.info(res.headers.get('accept')) // application/octet-stream
 const buf = await res.arrayBuffer() // raw bytes, not processed as JSON
 ```
 
+## Redirects
+
+If a requested URL contains a path component, that path component resolves to
+a UnixFS directory, but the URL does not have a trailing slash, one will be
+added to form a canonical URL for that resource, otherwise the request will
+be resolved as normal.
+
+```typescript
+import { verifiedFetch } from '@helia/verified-fetch'
+
+const res = await verifiedFetch('ipfs://bafyfoo/path/to/dir')
+
+console.info(res.url) // ipfs://bafyfoo/path/to/dir/
+```
+
+It's possible to prevent this behaviour and/or handle a redirect manually
+through use of the [redirect](https://developer.mozilla.org/en-US/docs/Web/API/fetch#redirect)
+option.
+
+## Example - Redirect: follow
+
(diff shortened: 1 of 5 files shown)


lidel (approved):
Thank you for wiring this up!

Option 4 sounds like a sensible compromise to implement fix for https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62, and will make things even easier in future when we accept HTTP URLs as-is. 

The `{ redirect: 'manual' }` is a very niche feature, and the fact we don't have opaque response should not have negative impact real world users.

The "spirit" of https://fetch.spec.whatwg.org/#atomic-http-redirect-handling  is to not leak cross-origin information not otherwise available, and we don't introduce any new info in `location` header or body (anyone can grab CID from URL and learn it is a directory themselves).

SgtPooki (approved):
Seems like a reasonable approach. I do wonder if modifying the response type with Object.defineProperty will mess with caching the response at all

PR #6, opened 2024-03-01
chore: unskip IPNS test

achingbrain:
This module supports retrieving raw IPNS records so the test for this can be unskipped.

## Change checklist


2color (approved):
🎉

PR #10, opened 2024-03-04
feat: support http range header

SgtPooki:
## Title

Add Range header support.

## Description

Fixes https://github.com/ipfs/helia-verified-fetch/issues/9

## Notes & open questions

- [x] Still need to add some additional tests, but it's ready for code review.
- [x] byte-range-context.spec.ts is currently skipping `handling ReadableStreams` tests, but these are covered by `range requests` tests.. going to try to fix quickly and if I can't i'll remove them.

## Change checklist


packages/verified-fetch/src/utils/responses.ts
```diff
@@ -40,7 +43,20 @@ export function okResponse (url: string, body?: BodyInit | null, init?: Response
   return response
 }
 
-export function notSupportedResponse (url: string, body?: BodyInit | null, init?: ResponseInit): Response {
+export function internalServerErrorResponse (url: string, body?: SupportedBodyTypes, init?: ResponseInit): Response {
```
SgtPooki, line 46: not sure if we want this to be `badInternalResponse` or other, so i went verbose

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -276,21 +277,50 @@ export class VerifiedFetch {
...
+        isRangeRequest = true
+      }
+    } catch (err) {
+      this.log.error('Error parsing range request', err)
+      return badRangeResponse('Invalid range request')
+    }
 
     try {
-      const pathDetails = await walkPath(this.helia.blockstore, `${cid.toString()}/${path}`, options)
+      const pathDetails = await walkPath(this.helia.blockstore, `${cid.toString()}/${path}`, { ...options, offset, length })
```
SgtPooki, line 304: I'm not sure if we care about offset and length here, but I presumed so. If not, we can remove a lot of this code up here and potentially remove the `eslint-disable-next-line complexity`

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -328,25 +358,71 @@ export class VerifiedFetch {
...
+      }
 
-    return response
+      return response
+    } catch (err) {
+      this.log.error('Error streaming %c/%s', cid, path, err)
+      if (isRangeRequest && (err as CodeError)?.code === 'ERR_INVALID_PARAMS') {
+        return badRangeResponse(resource)
+      }
+      return internalServerErrorResponse('Unable to stream content')
```
SgtPooki, line 424: returning an invalid response 500 response here resulted in needing to update `custom-dns-resolvers.spec.ts`. We may want to provide more detailed information than a 500 error here. Maybe a 503 "service unavailable" ?

packages/verified-fetch/test/range-requests.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    const content = await response.arrayBuffer()
+    expect(content.byteLength).to.equal(expected.byteSize) // the length of the data should match the requested range
+  }
+
+  async function assertFailingRange (response: Promise<Response>): Promise<void> {
+    await expect(response).to.eventually.have.property('status', 416)
+    await expect(response).to.eventually.have.property('statusText', 'Requested Range Not Satisfiable')
+  }
+
+  describe('unixfs', () => {
```
SgtPooki, line 66: I need to implement range-requests for 'raw' and add raw tests here

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -388,10 +464,35 @@ export class VerifiedFetch {
         this.log.error('Error parsing content type', err)
       }
     }
-
+    this.log.trace('setting content type to "%s"', contentType)
     response.headers.set('content-type', contentType)
   }
 
+  private checkForInvalidRangeRequest ({ offset, length, resource, stat }: { offset?: number, length?: number, resource: string, stat: UnixFSStats }): Response | undefined {
```
SgtPooki, line 471: I was able to move out some of the code from the handleDagPb, but could probably move this function outside of `verified-fetch` altogether.

SgtPooki (commented):
self review, some comments

packages/verified-fetch/src/utils/request-headers.ts
```diff
@@ -0,0 +1,86 @@
...
+export function getRequestRange (headers: Headers | HeadersInit | undefined, size?: bigint): HeliaRangeOptions | undefined {
+  const range = getHeader(headers, 'Range')
+  if (range == null) {
+    return undefined
+  }
+
+  /**
+   * Range: bytes=<start>-<end> | bytes=<start2>- | bytes=-<end2>
+   */
+  const match = range.match(/^bytes=((?<start>\d+)-(?<end>\d+)|(?<start2>\d+)-|-(?<end2>\d+))$/)
```
achingbrain, line 57: Don't know if we couldn't simplify this regex with something like:

```js
/^bytes=(?<start>\d+)?-(?<end>\d+)?$/
```

Then guard on neither `match.start` or `match.end` being present?

packages/verified-fetch/src/utils/responses.ts
achingbrain: I think call it what it is, 500 is `Internal Server Error` so that's what we should call it.

By that logic `notSupportedResponse` should also be `notImplementedResponse`, but whatevs.

packages/verified-fetch/src/utils/responses.ts
```diff
@@ -96,3 +112,63 @@ export function movedPermanentlyResponse (url: string, location: string, init?:
...
+ * TODO: Supporting multiple range requests will require additional changes to the `handleDagPb` and `handleRaw` functions in `src/verified-fetch.js`
+ */
+export function okRangeResponse (url: string, body: SupportedBodyTypes, range: RangeOptions, init?: ResponseOptions): Response {
+  // if we know the full size of the body, we should use it in the content-range header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
+  let contentRangeHeader: string | undefined
+
+  try {
+    contentRangeHeader = getContentRangeHeader({ body, total: range.totalSize, offset: range.offset, length: range.length })
+  } catch (e) {
+    return badRangeResponse(url, body, init)
```
achingbrain, line 134: It's probably worth logging the error to help ourselves in the future?

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -328,25 +358,71 @@ export class VerifiedFetch {
       }
     }
 
+    /**
+     * range-request validation. This shouldn't be required, but unixfs.cat isn't throwing in some cases.
```
achingbrain, line 362: Can you open an issue or better yet, a PR with some failing test cases for when `unixfs.cat` isn't throwing when it should?

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -328,25 +358,71 @@ export class VerifiedFetch {
...
-      response.headers.set('X-Ipfs-Roots', ipfsRoots.map(cid => cid.toV1().toString()).join(',')) // https://specs.ipfs.tech/http-gateways/path-gateway/#x-ipfs-roots-response-header
-    }
+      if (ipfsRoots != null) {
+        response.headers.set('X-Ipfs-Roots', ipfsRoots.map(cid => cid.toV1().toString()).join(',')) // https://specs.ipfs.tech/http-gateways/path-gateway/#x-ipfs-roots-response-header
+      }
 
-    return response
+      return response
+    } catch (err) {
+      this.log.error('Error streaming %c/%s', cid, path, err)
```
achingbrain, line 420: Just a tiny consistency thing, all of our errors are thrown with messages in sentence case, logging is all done in lower case.  Not sure why, but it's where we are.

```suggestion
      this.log.error('error streaming %c/%s', cid, path, err)
```

packages/verified-fetch/src/verified-fetch.ts
achingbrain: It could be made a function in the util folder?  No strong opinion.

packages/verified-fetch/src/utils/request-headers.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Range: bytes=<range-start>-
+ * Range: bytes=-<suffix-length> // must pass size so we can calculate the offset
+ *
+ * NOT SUPPORTED:
+ * Range: bytes=<range-start>-<range-end>, <range-start>-<range-end>
+ * Range: bytes=<range-start>-<range-end>, <range-start>-<range-end>, <range-start>-<range-end>
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range#directives
+ */
+export function getRequestRange (headers: Headers | HeadersInit | undefined, size?: bigint): HeliaRangeOptions | undefined {
```
achingbrain, line 48: Can we simplify the input types here and just make it `Headers`?

packages/verified-fetch/src/utils/request-headers.ts
```diff
@@ -0,0 +1,86 @@
+import { type CatOptions } from '@helia/unixfs'
+import { CodeError } from '@libp2p/interface'
+import { type ExporterOptions } from 'ipfs-unixfs-exporter'
+
+export function getHeader (headers: HeadersInit | undefined, header: string): string | undefined {
```
achingbrain, line 5: This is only used in the `getRequestRange` function, no need to export it.

If the input to `getRequestRange` is [simplified to Headers] this function can be removed entirely?

packages/verified-fetch/src/utils/response-headers.ts
```diff
@@ -0,0 +1,53 @@
+import { CodeError } from '@libp2p/interface'
+import type { SupportedBodyTypes } from '../types.js'
+
+/**
+ * Gets the body size of a given body if it's possible to calculate it synchronously.
+ */
+function syncBodySize (body: SupportedBodyTypes): number | null {
```
achingbrain, line 7: Minor nit but synchronous APIs normally have "sync" appended, not prepended.  For example the [fs sync API](https://nodejs.org/api/fs.html#synchronous-api).

```suggestion
function getBodySizeSync (body: SupportedBodyTypes): number | null {
```

achingbrain:
> ive got a refactor going that cleans this up a bunch

Done a first pass, will look again after the refactor lands.

packages/verified-fetch/src/utils/request-headers.ts
SgtPooki: yea I was debating doing that. I like the explicitness of what is already there, but I'm good with changing it. I'm sure having less alternatives will make it quicker. will give an attempt at this after I finish up the byte-range refactor

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: Yep. I wanted to make sure I was actually passing the right things.. i've got the test expectations and params cognizable now, so I'm sure something is off with .cat. will open issues when I can

packages/verified-fetch/test/custom-dns-resolvers.spec.ts
```diff
@@ -26,8 +26,9 @@ describe('custom dns-resolvers', () => {
       gateways: ['http://127.0.0.1:8080'],
       dnsResolvers: [customDnsResolver]
     })
-    // error of walking the CID/dag because we haven't actually added the block to the blockstore
-    await expect(fetch('ipns://some-non-cached-domain.com')).to.eventually.be.rejected.with.property('errors')
+    const response = await fetch('ipns://some-non-cached-domain.com')
+    expect(response.status).to.equal(500)
+    expect(response.statusText).to.equal('Internal Server Error')
```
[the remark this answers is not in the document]
SgtPooki: I felt the same about a 500 error. The closest error I could think of was 503, 409, 411 or some others depending on why exactly the streaming fails

packages/verified-fetch/src/utils/request-headers.ts
SgtPooki: I was trying to normalize the request Header processing and response header generation further with the existing code.

Unfortunately, the code I have now is already hard for *me* to adjust (and too many changes needed in too many places), and I wrote it, so I started creating a `byte-range-context.ts` file with a class `ByteRangeContext` that should be able to handle things more easily. When that's done, much of the byte-range handling should be much more straightforward.

Right now, calling getRequestRange, and getContentRangeHeader, and all the possible response values require too much context within handleDagPb & handleRaw, so I think this will get much better

packages/verified-fetch/src/utils/request-headers.ts
SgtPooki: I think we will continue to have a need for a getHeader function throughout this application though?

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: I talked with some folks earlier this week on this and it would make sense to default to 502 Bad Gateway

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: this function goes away with the addition of the byte-range-context class

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: fixed in [`f357a3d` (#10)](https://github.com/ipfs/helia-verified-fetch/pull/10/commits/f357a3d5eaa240a4808c8461cce65619fba2f1e4)

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: apparently it's working fine, it was a bug in my code :) surprised?! haha

packages/verified-fetch/test/custom-dns-resolvers.spec.ts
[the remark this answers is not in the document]
SgtPooki: discussed with @lidel the other day, kubo used to throw 502.. we can use that for now. fixed in [`83e80d8` (#10)](https://github.com/ipfs/helia-verified-fetch/pull/10/commits/83e80d85839cff64d5f9e8be308a0bf1d963382b)

packages/verified-fetch/src/utils/request-headers.ts
SgtPooki: function is removed and ByteRangeContext now handles parsing of header info upon construction.

packages/verified-fetch/src/utils/request-headers.ts
SgtPooki: resolving this as I think it makes sense for a normalized header getting function to exist in a request-headers utility file

packages/verified-fetch/src/utils/responses.ts
SgtPooki: done in [`05a6dfb` (#10)](https://github.com/ipfs/helia-verified-fetch/pull/10/commits/05a6dfb90631b527e078a4f9c61954c0a66e7905)

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: removed the offset & length here.

packages/verified-fetch/src/utils/response-headers.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * This function returns the value of the `Content-Range` header for a given range.
+ * If you know the total size of the body, you should pass it in the `options` object.
```
SgtPooki, line 3:
```suggestion
 * If you know the total size of the body, pass it as `byteSize`
```

packages/verified-fetch/src/utils/response-headers.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+  if (byteStart != null && byteEnd == null) {
+    return `bytes ${byteStart}-*/${total}`
+  }
+  if (byteStart == null && byteEnd != null) {
+    if (byteSize == null) {
+      return `bytes */${total}`
+    }
+    return `bytes ${byteSize - byteEnd + 1}-${byteSize}/${byteSize}`
+  }
```
SgtPooki, line 21:
```suggestion
  if (byteStart != null && byteEnd == null) {
    if (byteSize == null) {
      return `bytes */${total}`
    }
    return `bytes ${byteStart}-${byteSize}/${byteSize}`
  }
  if (byteStart == null && byteEnd != null) {
    if (byteSize == null) {
      return `bytes */${total}`
    }
    return `bytes ${byteSize - byteEnd + 1}-${byteSize}/${byteSize}`
  }
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -328,30 +332,58 @@ export class VerifiedFetch {
...
     }
 
+    if (byteRangeContext.isRangeRequest && byteRangeContext.isValidRangeRequest) {
+      stat = stat ?? await this.unixfs.stat(resolvedCID, {
+        signal: options?.signal,
+        onProgress: options?.onProgress
+      })
+      this.log.trace('stat for %c/%s: %o', resolvedCID, path, stat)
+      byteRangeContext.fileSize = stat.fileSize
+    }
```
SgtPooki, line 342:
```suggestion
    // we have to .stat before .cat so we can get the size and make sure byte offsets are calculated properly
    if (byteRangeContext.isRangeRequest && byteRangeContext.isValidRangeRequest) {
      stat = stat ?? await this.unixfs.stat(resolvedCID, {
        signal: options?.signal,
        onProgress: options?.onProgress
      })
      this.log.trace('stat.fileSize for rangeRequest %d', stat.fileSize)
      byteRangeContext.fileSize = stat.fileSize
    }
```

packages/verified-fetch/test/range-requests.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+    })
+  })
+
+  afterEach(async () => {
+    await stop(helia, verifiedFetch)
+  })
+
+  interface SuccessfulTestExpectation {
+    contentRange: string
+    // byteSize?: number
```
SgtPooki, line 32:
```suggestion
```

packages/verified-fetch/test/range-requests.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(response.statusText).to.equal('Partial Content')
+
+    expect(response).to.have.property('headers')
+    const contentRange = response.headers.get('content-range')
+    expect(contentRange).to.be.ok()
+    expect(contentRange).to.equal(expected.contentRange) // the response should include the range that was requested
+
+    const responseContent = await response.arrayBuffer()
+    expect(new Uint8Array(responseContent)).to.deep.equal(expected.bytes)
+    // expect(responseContent.byteLength).to.equal(expected.byteSize) // the length of the data should match the requested range
```
SgtPooki, line 52:
```suggestion
```

packages/verified-fetch/test/utils/byte-range-context.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+  validRanges.forEach(({ type, range, expected, body, contentRange }) => {
+    it(`should correctly slice ${type} body with range ${range}`, () => {
+      const context = new ByteRangeContext(logger, { Range: range })
+      context.setBody(body)
+      expect(context.getBody()).to.deep.equal(expected)
+      expect(context.contentRangeHeaderValue).to.equal(contentRange)
+    })
+  })
+
+  describe.skip('handling ReadableStreams', () => {
```
SgtPooki, line 100: currently investigating unskipping/removing these

SgtPooki (commented):
self review

SgtPooki (commented):
self review

packages/verified-fetch/test/utils/byte-range-context.spec.ts
SgtPooki: unskipped.

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,287 @@
...
+      this.log.trace('returning body with byteStart=%o, byteEnd=%o, byteSize=%o', byteStart, byteEnd, byteSize)
+      if (body instanceof ReadableStream) {
+        // stream should already be spliced by `unixfs.cat`
+        return body
+      }
+      return this.getSlicedBody(body)
+    }
+
+    // we should not reach this point, but return body untouched.
+    this.log.error('returning unmofified body for valid range request')
```
achingbrain, line 112:
```suggestion
    this.log.error('returning unmodified body for valid range request')
```

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,287 @@
...
+    this.byteStart = start
+    this.byteEnd = end
+    this.byteSize = byteSize
+  }
+
+  // This function returns the value of the "content-range" header.
+  // Content-Range: <unit> <range-start>-<range-end>/<size>
+  // Content-Range: <unit> <range-start>-<range-end>/*
+  // Content-Range: <unit> */<size>
+  // @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
```
achingbrain, line 271: If we're going to use jsdoc comments, the comment block needs to be `/** .... */` and not `//`

```suggestion
  /**
   * This function returns the value of the "content-range" header.
   * Content-Range: <unit> <range-start>-<range-end>/<size>
   * Content-Range: <unit> <range-start>-<range-end>/*
   * Content-Range: <unit> */<size>
   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
   */
```

packages/verified-fetch/src/utils/request-headers.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (headers == null) {
+    return undefined
+  }
+  if (headers instanceof Headers) {
+    return headers.get(header) ?? undefined
```
achingbrain, line 6: The `?? undefined` is redundant?

```suggestion
    return headers.get(header)
```

packages/verified-fetch/src/types.ts
```diff
@@ -1 +1,3 @@
 export type RequestFormatShorthand = 'raw' | 'car' | 'tar' | 'ipns-record' | 'dag-json' | 'dag-cbor' | 'json' | 'cbor'
+
+export type SupportedBodyTypes = string | ArrayBuffer | Blob | ReadableStream<Uint8Array> | null
```
achingbrain, line 3: Picking nits but `Types` is redundant in a type, `Supported` is a business-logic kind of decision not a type, so just `Body`?

```suggestion
export type Body = string | ArrayBuffer | Blob | ReadableStream<Uint8Array> | null
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -328,30 +332,59 @@ export class VerifiedFetch {
       }
     }
 
+    // we have to .stat before .cat so we can get the size and make sure byte offsets are calculated properly
+    if (byteRangeContext.isRangeRequest && byteRangeContext.isValidRangeRequest) {
+      stat = stat ?? await this.unixfs.stat(resolvedCID, {
```
achingbrain, line 337: This stat operation is unnecessary - you can disambiguate the `terminalElement` type by examining the `.type` field:

```js
if (terminalElement.type === 'file') {
  // tsc now knows terminalElement is `UnixFSFile`
  byteRangeContext.fileSize = terminalElement.unixfs.fileSize()
}

```

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    // we should not reach this point, but return body untouched.
+    this.log.error('returning unmofified body for valid range request')
+    return body
+  }
+
+  private getSlicedBody <T extends Uint8Array | ArrayBuffer | Blob | string>(body: T): T {
+    if (this.isPrefixLengthRequest) {
+      this.log.trace('sliced body with byteStart %o', this.byteStart)
+      return body.slice(this.offset) as T
```
achingbrain, line 119: All this casting removes any type safety we might have.  Would it be better to convert any passed body types to just one type instead of having to deal with all of: `Uint8Array | ArrayBuffer | Blob | string`?

Convert everything to `Blob` or something?

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,287 @@
...
+  public set fileSize (size: number | bigint | null) {
+    this._fileSize = size != null ? Number(size) : null
+    this.log.trace('set _fileSize to %o', this._fileSize)
+    // when fileSize changes, we need to recalculate the offset details
+    this.setOffsetDetails()
+  }
+
+  public get fileSize (): number | null | undefined {
+    return this._fileSize
+  }
```
achingbrain, line 150: If this class is just used internally, do we have to protect ourselves from ourselves with read-only properties like this?

packages/verified-fetch/src/utils/byte-range-context.ts
SgtPooki: +1 this was leftover from copying and pasting from MDN and doing an auto comment with vscode

packages/verified-fetch/src/types.ts
SgtPooki: I would rather do something like `ResponseBody`, but i'm good with any.

`Body` could easily be a type that comes from builtin/global types that could cause confusion. `SupportedBody` would be better I guess.. but it's explicitly informing devs reading the code that it's not just typical `Response.body` types.

SgtPooki (commented):
self review and response to comments

SgtPooki:
FYI: some tests were failing when merging main into byte-range PR due to new ipns things. changes at [`acdd632` (#10)](https://github.com/ipfs/helia-verified-fetch/pull/10/commits/acdd632437d9e5914e0d5a10adf1905c2e67ce04)

packages/verified-fetch/src/utils/byte-range-context.ts
achingbrain: Ok, it makes sense for `fileSize`, that change should probably be reverted.  What about `_isRangeRequest`?  It's a `readonly` boolean that's returned from a getter for `isRangeRequest`.

My concern with getters/setters is that they usually start off with good intentions around data access, but the temptation is always there to do business-logic things in them, which means accessing properties can mutate state which in the long term can make codebases very hard to reason about.

packages/verified-fetch/src/types.ts
achingbrain: `ResponseBody` would be fine.  It's a minor point tbh.

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,307 @@
...
+  private readonly _isRangeRequest: boolean
+
+  /**
+   * This property should only be set by calling `setFileSize` or `setBody`.
+   *
+   * @access private
+   */
+  public fileSize: Readonly<number | null | undefined>
+  private readonly _contentRangeHeaderValue: string | undefined
+  private _body: SupportedBodyTypes | null = null
```
achingbrain, line 53: `SupportedBodyTypes` includes `null`

```suggestion
  private _body: SupportedBodyTypes = null
```

achingbrain (approved):
LGTM, pending remaining discussion points

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,307 @@
...
+  /**
+   * This function returns the values of the "content-range" header.
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
+   *
+   * Returns data to support the following content ranges:
+   *
+   * @example
+   * - Content-Range: <unit> <byteStart>-<byteEnd>/<byteSize>
+   * - Content-Range: <unit> <byteStart>-<byteEnd>/&#8205;*
```
SgtPooki, line 289:
```suggestion
   * - Content-Range: <unit> <byteStart>-<byteEnd>/*
```

SgtPooki: I was testing jsdoc workarounds for js comment block closing tags and forgot to remove

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,307 @@
...
+   */
+  // - Content-Range: <unit> */<byteSize> // this is purposefully not in jsdoc block
```
SgtPooki, line 291: @achingbrain is this better than all as single line comments?

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -275,17 +277,20 @@ export class VerifiedFetch {
...
     } catch (err) {
-      this.log.error('Error walking path %s', path, err)
+      this.log.error('error walking path %s', path, err)
+
+      return badGatewayResponse('Error walking path')
     }
 
     let resolvedCID = terminalElement?.cid ?? cid
-    let stat: UnixFSStats
+    let stat: UnixFSStats | undefined
```
SgtPooki, line 293:
```suggestion
    let stat: UnixFSStats
```

Can probably remove this again since we only call stat once. If it requires no changes ill impl, otherwise ill leave it

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -323,30 +328,56 @@ export class VerifiedFetch {
       }
     }
 
+    // we have to .stat before .cat so we can get the size and make sure byte offsets are calculated properly
```
SgtPooki, line 331:
```suggestion
    // if terminalElement is a file, we know the size
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -408,7 +439,19 @@ export class VerifiedFetch {
...
+    let query: ParsedUrlStringResults['query']
+    try {
+      const result = await parseResource(resource, { ipns: this.ipns, logger: this.helia.logger }, options)
+      cid = result.cid
+      path = result.path
+      query = result.query
+    } catch (err) {
+      this.log.error('error parsing resource %s', resource, err)
+
+      return badRequestResponse('Invalid resource')
```
SgtPooki, line 453: Less throwing, more bad fetch responses

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  /**
+   * This property should only be set by calling `setFileSize` or `setBody`.
+   *
+   * @access private
+   */
+  public fileSize: Readonly<number | null | undefined>
+  private readonly _contentRangeHeaderValue: string | undefined
+  private _body: SupportedBodyTypes = null
+  private readonly _rangeRequestHeader: string | undefined
```
SgtPooki, line 54:
```suggestion
  private readonly rangeRequestHeader: string | undefined
```

packages/verified-fetch/src/utils/byte-range-context.ts
```diff
@@ -0,0 +1,307 @@
...
+  private byteStart: number | undefined
+  private byteEnd: number | undefined
+  private byteSize: number | undefined
+
+  constructor (logger: ComponentLogger, private readonly headers?: HeadersInit) {
+    this.log = logger.forComponent('helia:verified-fetch:byte-range-context')
+    this._rangeRequestHeader = getHeader(this.headers, 'Range')
+    if (this._rangeRequestHeader != null) {
+      this.log.trace('range request detected')
+      this._isRangeRequest = true
```
SgtPooki, line 68:
```suggestion
      this.isRangeRequest = true
```

SgtPooki (commented):
Going over code one last time

packages/verified-fetch/src/utils/byte-range-context.ts
SgtPooki: done

SgtPooki:
commenting elsewhere since the thread (https://github.com/ipfs/helia-verified-fetch/pull/10#discussion_r1521644018) was collapsed:

we could solve both concerns by setting internalFileSize so we know which value we set vs consumers.. but that could get confusing. i'll leave as is for now.

i've got _isRangeRequest updated locally, and have isValidRangeRequest updated as well.

So now we have:

inaccessable properties outside the ByteRangeContext class:

1. `_fileSize` // setter and getter
2. `_contentRangeHeaderValue` // getter (depends on other operations)
3. `_body` // set/get methods
4. ~~`_isValidRangeRequest` // getter (depends on other operations)~~

the only outlier here seems to be _body having `setBody` and `getBody` methods

---

inline update: `_isValidRangeRequest` has been removed. see [`d633456` (#10)](https://github.com/ipfs/helia-verified-fetch/pull/10/commits/d6334569dcf33ac63098ee8890ad3caab1114d3b)

PR #13, opened 2024-03-06
fix!: move dnsResolvers to options

2color:
## Description

As per https://github.com/ipfs/helia/pull/445#issuecomment-1980812623, this PR changes the API of `createVerifiedFetch` so as to allow passing a custom instance of Helia (for example, with a custom blockstore) in addition to customising the `dnsResolvers`. 

## Change checklist

### DIFF
--- a/packages/verified-fetch/README.md
+++ b/packages/verified-fetch/README.md
@@ -181,7 +181,8 @@ import { dnsJsonOverHttps, dnsOverHttps } from '@helia/ipns/dns-resolvers'
 
 const fetch = await createVerifiedFetch({
   gateways: ['https://trustless-gateway.link'],
-  routers: ['http://delegated-ipfs.dev'],
+  routers: ['http://delegated-ipfs.dev']
+}, {
   dnsResolvers: [
     dnsJsonOverHttps('https://my-dns-resolver.example.com/dns-json'),
     dnsOverHttps('https://my-dns-resolver.example.com/dns-query')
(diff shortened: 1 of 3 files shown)


achingbrain:
N.b. I renamed `chore!:` to `fix!:` because we want a release from this - `chore` will not cause one but `fix` will.

achingbrain:
I wonder if we could take this opportunity to make DNS resolvers more configurable in the stack.

Right now the other place we use DNS resolvers is in `@multiformats/multiaddr` to turn `/dns/example.com` into `/ip4/123.123.123.123`.  We do this in browsers through [hard-coded](https://github.com/vasco-santos/dns-over-http-resolver/blob/e81a1a879d509f319a70b14a9eae6971a1368a72/src/index.ts#L38-L39) HTTP-DNS servers which are globally modifiable 🤮 

I think we should:

1. Refactor `Multiaddr.resolve` to take a list of resolvers or a resolver object*
2. Use that in preference to the global list if passed
3. Add it to the libp2p config and pass it/them to `ma.resolve` during dialing
4. Add it to the Helia config and make it available as part of the Helia interface
5. Refactor `@helia/ipns` to use the DNS resolvers from the Helia interface

\* = a resolver object rather than a simple list might give us a bit of flexibility to accommodate things like [tld-specific resolvers](https://github.com/ipfs/kubo/blob/master/docs/config.md#dns).

2color:
Closing in favour of https://github.com/ipfs/helia-verified-fetch/pull/18

PR #19, opened 2024-03-13
fix: set cache-control header correctly

### LINKED ISSUE #17: Cache-Control headers for mutable resources are wrong
## Summary

When requesting resources with the `ipns://` prefix, verifiedFetch returns the wrong `Cache-Control` header (specifically `'public, max-age=29030400, immutable'`) in the response object according to [the specs](https://specs.ipfs.tech/http-gateways/path-gateway/#cache-control-response-header). 

This happens for both IPNS names and DNSLink

## Suggestion

Set the `Cache-Control` headers according to [the specs](https://specs.ipfs.tech/http-gateways/path-gateway/#cache-control-response-header).

2color:
## Description

Set the cache-control cache header conditionally for immutable responses.

For mutable responses, i.e. IPNS and DNSLink, the correct `max-age` in the cache-control would be the TTL of the record . Determining the right TTL would require modifying the `parseUrlString` method and a few of the methods to also return TTL and is outside the scope of this PR. Should be ok initially to just to omit the header given that Kubo does this.

fixes #17
## Change checklist


2color:
As discussed in Helia WG:

Until we change the APIs to return the TTL for DNSLink and IPNS records, we should default to a 5 minute `max-age`. This is in line with Kubo (see https://github.com/ipfs/boxo/issues/329#issuecomment-1995236409)

achingbrain:
Related - https://github.com/ipfs/helia/pull/471 - exposes the resolved IPNS record and DNSLink Answers to IPNS callers to allow accurately setting this field.

SgtPooki:
i'll get this merged in shortly.

SgtPooki (approved):
self review.

SgtPooki:
@2color if you're good with the changes here, we should merge it.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -23,30 +23,66 @@ export interface ParsedUrlQuery extends Record<string, string | unknown> {
...
+ * @see https://github.com/ipfs/js-ipns/blob/16e0e10682fa9a663e0bb493a44d3e99a5200944/src/index.ts#L200
+ * @see https://github.com/ipfs/js-ipns/pull/308
+ */
+function calculateTtl (resolveResult?: IPNSResolveResult | DNSLinkResolveResult): number | undefined {
+  if (resolveResult == null) {
+    return undefined
+  }
+  const dnsLinkTtl = (resolveResult as DNSLinkResolveResult).answer?.TTL
+  const ipnsTtlNs = (resolveResult as IPNSResolveResult).record?.ttl
+  // For some reason, ipns "nanoseconds" are 1e-8 of a second, instead of 1e-9.
```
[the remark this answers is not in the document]
2color: @hacdias Any chance you can give this a look. I'm too confused

packages/verified-fetch/src/utils/response-headers.ts
```diff
@@ -1,3 +1,41 @@
...
+     * default limit for unknown TTL: "use 5 minute as default fallback when it is not available."
+     *
+     * @see https://github.com/ipfs/boxo/issues/329#issuecomment-1995236409
+     */
+    headerValue = 'public, max-age=300'
+  } else {
+    headerValue = `public, max-age=${ttl}`
+  }
+
+  if (headerValue != null) {
```
2color, line 34: nit: no need for this check since the last `else` clause catching all possible values, `headerValue` will always be set to something.

packages/verified-fetch/test/cache-control-header.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    const resp = await verifiedFetch.fetch(`ipns://${peerId}`)
+    expect(resp).to.be.ok()
+    expect(resp.status).to.equal(200)
+
+    expect(resp.headers.get('Cache-Control')).to.equal(`public, max-age=${oneHourInSeconds}`)
+  })
+
+  it('should not contain immutable in the cache-control header for a DNSLink name', async () => {
+    verifiedFetch = new VerifiedFetch({
```
2color, line 112: The code above (https://github.com/ipfs/helia-verified-fetch/pull/19/files#diff-853a9f8afb602ec83fff59dd64982c5e2adf008e30cdcf9c179fe361f0455c48R32-R43) already instantiates verifiedFetch with the customDnsResolver, so it should be needed here.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -23,30 +23,66 @@ export interface ParsedUrlQuery extends Record<string, string | unknown> {
...
+  queryString?: string
+
+}
+function matchURLString (urlString: string): MatchUrlGroups {
   for (const pattern of [URL_REGEX, PATH_REGEX, PATH_GATEWAY_REGEX, SUBDOMAIN_GATEWAY_REGEX]) {
     const match = urlString.match(pattern)
 
     if (match?.groups != null) {
-      return match.groups
+      return match.groups as unknown as MatchUrlGroups // force cast to MatchUrlGroups, because if it matches, it has to contain this structure.
```
2color, line 55: Do you know why TS require you to first cast to unknown?

packages/verified-fetch/src/utils/parse-url-string.ts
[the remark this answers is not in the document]
hacdias: @2color that does seem like a bug indeed, the tests also use the wrong conversion it seems like: https://github.com/ipfs/js-ipns/blob/main/test/index.spec.ts

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: its a regex match which doesn't support great generics AFAIU?

packages/verified-fetch/test/cache-control-header.spec.ts
SgtPooki: verifiedFetch in `beforeEach` doesn't pass dnsResolvers, so this fails without it.

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: https://github.com/microsoft/TypeScript/issues/32098

packages/verified-fetch/src/utils/response-headers.ts
SgtPooki: I think typescript was yelling about this, but could be a leftover bug from some refactoring.

packages/verified-fetch/src/utils/parse-url-string.ts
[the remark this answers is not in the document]
SgtPooki: https://github.com/ipfs/js-ipns/commit/e4175ccf887d8ebc5590693759ef46b31a5ee18f#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556 was the commit that changed from `100000n` to `BigInt(100000)`

https://github.com/ipfs/js-ipns/commit/d1421f9389961ac6bf7b0ab4d80442fd7b78c14f#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556 changed from `new NanoDate(Date.now() + Number(lifetime)).toString()` to the broken `100000n` value for lifetimeNs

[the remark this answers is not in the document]
SgtPooki: discovered with:

1. `git log -S 'BigInt(100000)'` - find first commit (last in list).. see that `10000n` was prev value
2. `git log -S '100000n'` - find last commit, see that `d1421f9389961ac6bf7b0ab4d80442fd7b78c14f` was the last first one to add `10000n`

@rvagg @achingbrain do we know why js-ipns seems to be calculating nanoseconds incorrectly?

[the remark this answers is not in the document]
SgtPooki: FYI I opened up https://github.com/ipfs/js-ipns/issues/310 which clarifies the problem more specifically.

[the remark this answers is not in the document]
SgtPooki: FYI @2color this is tested, and will break when js-ipns is fixed (should release as a breaking change), which will cause the tests here to fail when we update.

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: also: casting to unknown is a way to get typescript to remove all type inferences.. so that we can explicitly cast as `MatchUrlGroups`. The key to verify here, since we're explicitly casting the type, is that the `MatchUrlGroups` is correct. Are things that "MUST" match marked as required? are things that may not match marked as optional?

SgtPooki: We could also do a type guard, but thats no more typesafe IMHO

packages/verified-fetch/test/cache-control-header.spec.ts
2color: It has the Helia instance passed to it with the customDnsResolver 

https://github.com/ipfs/helia-verified-fetch/pull/19/files#diff-853a9f8afb602ec83fff59dd64982c5e2adf008e30cdcf9c179fe361f0455c48R36

SgtPooki: you're right.. removing

packages/verified-fetch/src/utils/response-headers.ts
SgtPooki: was able to remove this and things still passed. makes sense.

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: i'm adding a typeguard.. it's annoying but it *is* more type-safe.

Issue #25, opened 2024-03-19
Abort signal is not respected


2color:
When passing an `new AbortController().signal` to a verifiedFetch request, aborting does not seem to work. 

You can see this in action in the reproduction, where if you click to fetch IPNS and right after click Abort Fetch, the `controller.abort()` call does not cause the promise from the `verifiedFetch` call to throw an error as [you'd expect in `Fetch`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController#examples)

To make this easier to test, I've added a reproduction:

[...]

PR #26, opened 2024-03-21
feat: abort signals are respected

SgtPooki:
## Title

feat: abort signals are respected

## Description

fixes https://github.com/ipfs/helia-verified-fetch/issues/25

need some eyes on this to ensure the approach is sound and that we're exiting as early as possible.

## Notes & open questions

slightly WIP but I do need feedback, so PR not marked as WIP.

There was an obvious bug where signal was not being passed through fully. :

1. no signal being read at all (it was always undefined): see the changes to `convertOptions`
2. parse-resource did not previously accept a signal.
2. parse-url-string did not previously accept a signal.

I'm happy to revert some of the explicit signal handling and just get through the fixes to the above, but I think we should exit early when we can.

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
+++ b/packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
@@ -5,7 +5,7 @@ import type { ComponentLogger } from '@libp2p/interface'
 /**
  * Converts an async iterator of Uint8Array bytes to a stream and returns the first chunk of bytes
  */
-export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8Array>, path: string, logger: ComponentLogger, options?: Pick<VerifiedFetchInit, 'onProgress'>): Promise<{ stream: ReadableStream<Uint8Array>, firstChunk: Uint8Array }> {
+export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8Array>, path: string, logger: ComponentLogger, options?: Pick<VerifiedFetchInit, 'onProgress' | 'signal'>): Promise<{ stream: ReadableStream<Uint8Array>, firstChunk: Uint8Array }> {
   const log = logger.forComponent('helia:verified-fetch:get-stream-from-async-iterable')
   const reader = iterator[Symbol.asyncIterator]()
   const { value: firstChunk, done } = await reader.next()
--- a/packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
+++ b/packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
@@ -23,6 +23,11 @@ export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8A
     },
     async pull (controller) {
       const { value, done } = await reader.next()
+      if (options?.signal?.aborted === true) {
+        controller.error(new Error(options.signal.reason ?? 'signal aborted by user'))
+        controller.close()
+        return
+      }
 
       if (done === true) {
         if (value != null) {
--- a/packages/verified-fetch/src/utils/parse-resource.ts
+++ b/packages/verified-fetch/src/utils/parse-resource.ts
@@ -3,7 +3,7 @@ import { parseUrlString } from './parse-url-string.js'
 import type { ParsedUrlStringResults } from './parse-url-string.js'
 import type { Resource } from '../index.js'
 import type { IPNS, IPNSRoutingEvents, ResolveDNSLinkProgressEvents, ResolveProgressEvents } from '@helia/ipns'
-import type { ComponentLogger } from '@libp2p/interface'
+import type { AbortOptions, ComponentLogger } from '@libp2p/interface'
 import type { ProgressOptions } from 'progress-events'
 
 export interface ParseResourceComponents {
@@ -11,7 +11,7 @@ export interface ParseResourceComponents {
   logger: ComponentLogger
 }
 
-export interface ParseResourceOptions extends ProgressOptions<ResolveProgressEvents | IPNSRoutingEvents | ResolveDNSLinkProgressEvents> {
+export interface ParseResourceOptions extends ProgressOptions<ResolveProgressEvents | IPNSRoutingEvents | ResolveDNSLinkProgressEvents>, AbortOptions {
 
 }
 /**
@@ -21,7 +21,7 @@ export interface ParseResourceOptions extends ProgressOptions<ResolveProgressEve
  */
 export async function parseResource (resource: Resource, { ipns, logger }: ParseResourceComponents, options?: ParseResourceOptions): Promise<ParsedUrlStringResults> {
   if (typeof resource === 'string') {
-    return parseUrlString({ urlString: resource, ipns, logger }, { onProgress: options?.onProgress })
+    return parseUrlString({ urlString: resource, ipns, logger }, options)
   }
 
   const cid = CID.asCID(resource)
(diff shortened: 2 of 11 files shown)


packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -147,6 +149,8 @@ export async function parseUrlString ({ urlString, ipns, logger }: ParseUrlStrin
     }
   }
 
+  options?.signal?.throwIfAborted()
```
SgtPooki, line 152: this could probably be moved into `verified-fetch.ts` `fetch` method.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -283,10 +286,18 @@ export class VerifiedFetch {
...
+    } catch (err: any) {
       this.log.error('error walking path %s', path, err)
+      /**
+       * TODO: do this better. We should be able to distinguish between an abortError and a regular error
+       * However, `helia:networked-storage:error` throws an AggregateError with the abort error inside it
+       */
+      const abortError = err.errors?.find((e: Error) => e.message.includes('abort'))
+      if (abortError != null) {
+        return badRequestResponse(resource.toString(), abortError.message)
+      }
```
SgtPooki, line 298: this is the only place where i'm not sure we want to keep the code from this PR

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -430,11 +442,35 @@ export class VerifiedFetch {
     [identity.code]: this.handleRaw
   }
 
+  /**
+   *
+   * TODO: Should we use 400, 408, 418, or 425, or throw and not even return a response?
```
SgtPooki, line 447: we should discuss this.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -430,11 +442,35 @@ export class VerifiedFetch {
...
+   */
+  private async abortHandler (opController: AbortController): Promise<void> {
+    this.log.error('signal aborted by user')
+    opController.abort('signal aborted by user')
+    await this.stop()
```
SgtPooki, line 452: we should discuss this as well. without calling stop, sometimes things don't quit. bug in Helia? should we call `.start()` immediately after?

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -430,11 +442,35 @@ export class VerifiedFetch {
...
     this.log('fetch %s', resource)
 
     const options = convertOptions(opts)
 
+    const opController = new AbortController()
+    if (options?.signal != null) {
+      options.signal.onabort = this.abortHandler.bind(this, opController)
+      options.signal = opController.signal
+    }
+    this.log('options %o', options)
```
SgtPooki, line 472:
```suggestion
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -430,11 +442,35 @@ export class VerifiedFetch {
...
+   * always exit early (and cleanly) if a given signal is aborted
+   */
   async fetch (resource: Resource, opts?: VerifiedFetchOptions): Promise<Response> {
     this.log('fetch %s', resource)
 
     const options = convertOptions(opts)
 
+    const opController = new AbortController()
+    if (options?.signal != null) {
+      options.signal.onabort = this.abortHandler.bind(this, opController)
```
SgtPooki, line 469: we override any handlers they have on their signal.. they should wrap any signals they want to provide

packages/verified-fetch/test/fixtures/dns-answer-fake.ts
```diff
@@ -0,0 +1,13 @@
+import { stubInterface } from 'sinon-ts'
+import type { DNSResponse } from '@multiformats/dns'
+
+export function answerFake (data: string, TTL: number, name: string, type: number): DNSResponse {
```
SgtPooki, line 4: i believe this is in the https://github.com/ipfs/helia-verified-fetch/pull/19 PR, and i needed it here.

SgtPooki (commented):
self review

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -78,7 +78,10 @@ function convertOptions (options?: VerifiedFetchOptions): (Omit<VerifiedFetchOpt
   let signal: AbortSignal | undefined
   if (options?.signal === null) {
     signal = undefined
+  } else {
+    signal = options?.signal
```
SgtPooki, line 82: the main part that fixes the bug mentioned by #25

packages/verified-fetch/src/utils/parse-resource.ts
```diff
@@ -1,17 +1,16 @@
...
 import type { ComponentLogger } from '@libp2p/interface'
-import type { ProgressOptions } from 'progress-events'
 
 export interface ParseResourceComponents {
   ipns: IPNS
   logger: ComponentLogger
 }
 
-export interface ParseResourceOptions extends ProgressOptions<ResolveProgressEvents | IPNSRoutingEvents | ResolveDNSLinkProgressEvents> {
+export interface ParseResourceOptions extends ParseUrlStringOptions {
```
SgtPooki, line 13: we extend options accepted by ParseUrlStringOptions

packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
```diff
@@ -5,7 +5,7 @@ import type { ComponentLogger } from '@libp2p/interface'
 /**
  * Converts an async iterator of Uint8Array bytes to a stream and returns the first chunk of bytes
  */
-export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8Array>, path: string, logger: ComponentLogger, options?: Pick<VerifiedFetchInit, 'onProgress'>): Promise<{ stream: ReadableStream<Uint8Array>, firstChunk: Uint8Array }> {
+export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8Array>, path: string, logger: ComponentLogger, options?: Pick<VerifiedFetchInit, 'onProgress' | 'signal'>): Promise<{ stream: ReadableStream<Uint8Array>, firstChunk: Uint8Array }> {
```
SgtPooki, line 8: another part of the bug fix... we need to be passing progressOptions and signal through. I just realized I wasn't checking for signal abort here... the iterator should handle this.. but i'll update just in case.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -13,7 +13,7 @@ export interface ParseUrlStringInput {
   ipns: IPNS
   logger: ComponentLogger
 }
-export interface ParseUrlStringOptions extends ProgressOptions<ResolveDNSLinkProgressEvents> {
+export interface ParseUrlStringOptions extends ProgressOptions<ResolveProgressEvents | IPNSRoutingEvents | ResolveDNSLinkProgressEvents>, AbortOptions {
```
SgtPooki, line 16: another part of the bug fix. we should accept a `{signal}`

SgtPooki (commented):
self review.. maybe some duplicated comments.

packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
```diff
@@ -23,6 +23,11 @@ export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8A
     },
     async pull (controller) {
       const { value, done } = await reader.next()
+      if (options?.signal?.aborted === true) {
+        controller.error(new Error(options.signal.reason ?? 'signal aborted by user'))
+        controller.close()
+        return
+      }
```
SgtPooki, line 30: added in [`3b58fa0` (#26)](https://github.com/ipfs/helia-verified-fetch/pull/26/commits/3b58fa0bf5502a9e2868866a0be8accba4efc2e7).

iterator from `unixfs.cat` should handle failing this stream, but handling this explicitly may allow us to exit earlier

SgtPooki (commented):
reviewed last changes. will merge shortly

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: removed calling of `this.stop()`

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: we can change this later if we want.. but it would be good to be aware of this. @achingbrain

Issue #31, opened 2024-03-25
feat: tree-shake ipfs-unixfs-importer


SgtPooki:
since we're not using any importing functionality in verified-fetch, we should make sure our resulting bundle does not include `ipfs-unixfs-importer` (or others.. which can be handled in separate issue)

running `npm why ipfs-unixfs-importer` in helia-service-worker-gateway shows the following:

```
ipfs-unixfs-importer@15.2.4
node_modules/ipfs-unixfs-importer
  ipfs-unixfs-importer@"^15.2.4" from @helia/unixfs@3.0.2
  node_modules/@helia/unixfs
    @helia/unixfs@"^3.0.1" from @helia/verified-fetch@1.3.0
    node_modules/@helia/verified-fetch
      @helia/verified-fetch@"^1.3.0" from the root project
```

Can we explicitly remove ipfs-unixfs-importer during esbuild, or do we need to modify @helia/unixfs to export a "getter only" version?

SgtPooki:
### with regular build:

```bash
╰─ ✔ ❯ npm run build -- -b

> @helia/verified-fetch@1.3.1 build
> aegir build -b

[10:41:55] tsc [started]
[10:41:56] tsc [completed]
[10:41:56] esbuild [started]
[10:41:56] → Use https://esbuild.github.io/analyze/ to load "./dist/stats.json".
[10:41:56] → 124.57KB (▼-7.43KB / 132KB)
[10:41:56] esbuild [completed]
```

<img width="1282" alt="image" src="https://github.com/ipfs/helia-verified-fetch/assets/1173416/f1578944-7433-4682-b279-4b61bed66aa9">


### with `ipfs-unixfs-importer` marked as `external`

```bash
╰─ ✔ ❯ npm run build -- -b

> @helia/verified-fetch@1.3.1 build
> aegir build -b

[10:44:25] tsc [started]
[10:44:26] tsc [completed]
[10:44:26] esbuild [started]
[10:44:27] → Use https://esbuild.github.io/analyze/ to load "./dist/stats.json".
[10:44:27] → 116.04KB (▼-15.96KB / 132KB)
[10:44:27] esbuild [completed]
```

.aegir.js changes:

```diff
diff --git a/packages/verified-fetch/.aegir.js b/packages/verified-fetch/.aegir.js
index 89b5633..cbd9809 100644
--- a/packages/verified-fetch/.aegir.js
+++ b/packages/verified-fetch/.aegir.js
[...]

PR #32, opened 2024-03-25
chore: remove ipfs-unixfs-importer from bundle

SgtPooki:
## Title

chore: remove ipfs-unixfs-importer from bundle

## Description

fixes https://github.com/ipfs/helia-verified-fetch/issues/31

## Notes & open questions

removes ipfs-unixfs-importer from the bundle

This is potentially dangerous because we're explicitly removing it from the resulting bundle and ipfs-unixfs-importer may export some things that are needed by @helia/unixfs for non-importing cases.

However, tests seem to pass locally with this change.

## Change checklist

### DIFF
--- a/packages/verified-fetch/.aegir.js
+++ b/packages/verified-fetch/.aegir.js
@@ -1,8 +1,13 @@
 /** @type {import('aegir').PartialOptions} */
 const options = {
   build: {
-    bundlesizeMax: '132KB'
-  }
+    bundlesizeMax: '132KB',
+    config: {
+      external: [
+        'ipfs-unixfs-importer'
+      ],
+    }
+  },
 }
 
 export default options


2color (approved):

2color:
Note to self: 

This is affecting the `external` config of esbuild https://esbuild.github.io/api/#external reference : https://github.com/ipfs/aegir/blob/dc5c226c51a592d8e7a84a7143154b1a4da78097/src/types.ts#L64

achingbrain:
I've opened https://github.com/ipfs/helia-verified-fetch/pull/42 as an alternative here.  Instead of remove deps from the tree (which may have unpredictable results), we could just remove `@helia/unixfs` and use the exporter directly, since we only stat and cat a DAG we don't need all the code that deals with writing DAGs.

SgtPooki:
we should follow the approach from #42

PR #33, opened 2024-03-25
fix: unixfs dir redirect

SgtPooki:
- fix: check redirect trailing slash if path is empty
- test: add interop test for unixfs directory redirect

## Title

fix: unixfs dir redirect

## Description

Addresses issue found while working on https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62 where the redirect for a directory was not being followed if the path was empty.

* https://github.com/ipfs/helia-verified-fetch/pull/5 added redirect handling for unixfs directories, but did not account for the case where the path was empty.
* Previous code in https://github.com/ipfs/helia-verified-fetch/pull/15 added tests for a subdirectory redirect, but did not test the case where the path was empty (which is the case when the root CID requested is a directory).

## Change checklist


SgtPooki (commented):
self review

SgtPooki (commented):
self review latest changes

PR #34, opened 2024-03-26
fix: ttl and caching for ipns urls

2color:
## Description

This PR mainly fixes the TLRU cache behaviour by ensuring that when setting DNSLink and IPNS resolution answers to the cache, **we correctly convert the TTL from seconds to milliseconds as expected by the TLRU cache** 

In addition: 
- Adds tests to ensure that the correct TTL is returned by the `parseUrlString` function. 
- Converts the IPNS record TTL from nanoseconds to seconds correctly. 

## Notes & open questions

- This should fix some of the duplicate resolutions reported in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/59. However, it should 
- We still need to follow up in js-ipns regarding why [nanoseconds are incorrectly calculated](https://github.com/ipfs/js-ipns/blob/e933496d93e553662b968019b57a2f8413fbefc6/src/index.ts#L200) when converting from ms. 

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/utils/parse-url-string.ts
+++ b/packages/verified-fetch/src/utils/parse-url-string.ts
@@ -214,11 +214,14 @@ export async function parseUrlString ({ urlString, ipns, logger }: ParseUrlStrin
     throw new AggregateError(errors, `Invalid resource. Cannot determine CID from URL "${urlString}"`)
   }
 
-  const ttl = calculateTtl(resolveResult)
+  let ttl = calculateTtl(resolveResult)
 
   if (resolveResult != null) {
     // use the ttl for the resolved resouce for the cache, but fallback to 2 minutes if not available
-    ipnsCache.set(cidOrPeerIdOrDnsLink, resolveResult, ttl ?? 60 * 1000 * 2)
+    ttl = ttl ?? 60 * 2
+    log.trace('caching %s resolved to %s with TTL: %s', cidOrPeerIdOrDnsLink, cid, ttl)
+    // convert ttl from seconds to ms for the cache
+    ipnsCache.set(cidOrPeerIdOrDnsLink, resolveResult, ttl * 1000)
   }
 
   // parse query string
--- a/packages/verified-fetch/src/utils/parse-url-string.ts
+++ b/packages/verified-fetch/src/utils/parse-url-string.ts
@@ -85,6 +85,7 @@ function matchURLString (urlString: string): MatchUrlGroups {
  *
  * @see https://github.com/ipfs/js-ipns/blob/16e0e10682fa9a663e0bb493a44d3e99a5200944/src/index.ts#L200
  * @see https://github.com/ipfs/js-ipns/pull/308
+ * @returns the ttl in seconds
  */
 function calculateTtl (resolveResult?: IPNSResolveResult | DNSLinkResolveResult): number | undefined {
   if (resolveResult == null) {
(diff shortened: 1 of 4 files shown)


lidel (changes requested):

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -85,15 +85,15 @@ function matchURLString (urlString: string): MatchUrlGroups {
...
  */
 function calculateTtl (resolveResult?: IPNSResolveResult | DNSLinkResolveResult): number | undefined {
   if (resolveResult == null) {
     return undefined
   }
   const dnsLinkTtl = (resolveResult as DNSLinkResolveResult).answer?.TTL
   const ipnsTtlNs = (resolveResult as IPNSResolveResult).record?.ttl
-  // For some reason, ipns "nanoseconds" are 1e-8 of a second, instead of 1e-9.
-  const ipnsTtl = ipnsTtlNs != null ? Number(ipnsTtlNs / BigInt(1e8)) : undefined
+  const ipnsTtl = ipnsTtlNs != null ? Number(ipnsTtlNs / BigInt(1e9)) : undefined
```
SgtPooki, line 96: I feel like we need a check for js-ipns ie8 nanoseconds here.

packages/verified-fetch/test/cache-control-header.spec.ts
```diff
@@ -71,7 +71,10 @@ describe('cache-control header', () => {
     expect(resp.headers.get('Cache-Control')).to.not.containIgnoreCase('immutable')
   })
 
-  it('should return the correct max-age in the cache-control header for an IPNS name', async () => {
+  // Skipping until https://github.com/ipfs/js-ipns/issues/310 is resolved
+  // Note that the source of the error is from the `name.publish` call rather than the max-age value
```
SgtPooki, line 75: 👍

packages/verified-fetch/test/utils/parse-url-string.spec.ts
```diff
@@ -257,6 +257,40 @@ describe('parseUrlString', () => {
...
+      const testPeerId = await createEd25519PeerId()
+
+      ipns.resolve.withArgs(matchPeerId(testPeerId)).resolves({
+        cid: CID.parse('QmQJ8fxavY54CUsxMSx9aE9Rdcmvhx8awJK2jzJp4iAqCr'),
+        path: '',
+        record: ipnsRecordStub({ peerId: testPeerId, ttl: oneHourInNanoseconds })
+      })
+
+      const result = await parseUrlString({ urlString: `ipns://${testPeerId}`, ipns, logger })
+      expect(result.ttl).to.equal(3600)
```
SgtPooki, line 290:
```suggestion
      expect(result.ttl).to.equal(oneHourInSeconds)
```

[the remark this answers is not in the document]
2color: > I am afraid generating record stub in JS and testing it with JS does not give us enough coverage for this to be future-proof.

Me neither 😅 


> @2color do you mind refactoring this test to use the static IPNS record from gateway-conformance? (it is also ok to keep this as-is, and add this upstream in `@helia/ipns`, but we need to have it):

Yeah I think that test belongs in `@helia/ipns`. Here we're  just testing that the TTL returned from `@helia/ipns` for IPNS and DSLLink is returned in the correct unit (seconds).

SgtPooki (approved):

packages/verified-fetch/test/utils/parse-url-string.spec.ts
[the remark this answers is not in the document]
SgtPooki: This test does confirm that given a particular ttl from a record, our parsing of the TTL is correct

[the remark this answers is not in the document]
SgtPooki: A test of the ipnsrecord as lidel suggests should go in interop tests

packages/verified-fetch/src/utils/parse-url-string.ts
2color: What do you mean? 

here's no way for us to know whether the TTL was created with an buggy version which is off by an order of magnitude (1e8)

SgtPooki: Yea i know... We need to fix js-ipns

packages/verified-fetch/test/utils/parse-url-string.spec.ts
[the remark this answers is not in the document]
2color: So I checked and we have exactly a test that implicitly checks this by ensuring js-ipns can roundtrip marshalling a record created by Kubo: https://github.com/ipfs/js-ipns/blob/main/test/index.spec.ts#L402-L414 

I also opened https://github.com/ipfs/helia/pull/476 as an additional check.  

Any objection on merging this @lidel?

lidel (approved):

Issue #35, opened 2024-03-30
bug: gateway-conformance support for query parameter


SgtPooki:
i'm doing some tests on https://github.com/ipfs/helia-http-gateway/pull/67/ and noticed one oddity. we're not converting dag-json properly? the content-type is being set, incorrectly, to `text/plain; charset=utf-8`

We should be falling into the block similar to https://github.com/ipfs/helia-verified-fetch/blob/39bb14e0a7a8441ff0c323db15264435e0121b23/packages/verified-fetch/src/verified-fetch.ts#L243-L249 but it doesn't look like we have a path from `raw` to `dag-json` ?

requesting http://helia-http-gateway.localhost/ipfs/bafkreialihlqnf5uwo4byh4n3cmwlntwqzxxs2fg5vanqdi3d7tb2l5xkm?format=dag-json results in the following output from verified-fetch:

```
[helia] 2024-03-30T00:01:23.669Z helia:verified-fetch fetch http://helia-http-gateway.localhost/ipfs/bafkreialihlqnf5uwo4byh4n3cmwlntwqzxxs2fg5vanqdi3d7tb2l5xkm?format=dag-json
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch incoming accept header "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8"
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch incoming query format "dag-json", mapped to application/vnd.ipld.dag-json
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch output type application/octet-stream
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch:trace finding handler for cid code "85" and output type "application/octet-stream"
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch:trace calling handler "handleRaw"
[helia] 2024-03-30T00:01:23.676Z helia:verified-fetch:byte-range-context:trace no range request detected
[helia] 2024-03-30T00:01:23.677Z helia:trustless-gateway-block-broker getting block for bafkreialihlqnf5uwo4byh4n3cmwlntwqzxxs2fg5vanqdi3d7tb2l5xkm from http://127.0.0.1:8081/
[...]

PR #36, opened 2024-03-30
fix: implicit accept header can be overridden by format query

SgtPooki:
## Title

fix: implicit accept header can be overridden by format query

## Description

Fixes https://github.com/ipfs/helia-verified-fetch/issues/35 & also allows mapping of 'raw' to 'dag-json'. I imagine users will get errors if they try to accept dag-json for non-json content, we can address that later?

## Notes & open questions

Basically, we were previously doing the following:

1. check accept header
2. check format query param
3. if accept header exists, use it, otherwise, use the mapped query param format

However, that was causing the bug in https://github.com/ipfs/helia-verified-fetch/issues/35 because the format query param was being ignored if the accept header was present (which it usually is for most clients), even if it wasn't [a valid IPFS gateway Accept header](https://specs.ipfs.tech/http-gateways/path-gateway/#accept-request-header)

Now, we do this:

1. check accept header
2. check format query param
4. if accept header is not a valid IPFS gateway accept header, and we have a mapping for the format query param, use the mapped query param format value.
3. otherwise use the accept header

more notes are in https://github.com/ipfs/helia-verified-fetch/issues/35

## Change checklist


SgtPooki (commented):
self review

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -710,4 +710,63 @@ describe('@helia/verifed-fetch', () => {
...
+      const obj = {
+        hello: 'world'
+      }
+      const c = dagCbor(helia)
+      const cid = await c.add(obj)
+
+      const resp = await verifiedFetch.fetch(`http://example.com/ipfs/${cid}?format=dag-json`, {
+        headers: {
+          // see https://github.com/ipfs/helia-verified-fetch/issues/35
+          // accept: '*/*'
```
achingbrain, line 743: Is this supposed to be commented out?

packages/verified-fetch/src/utils/is-accept-explicit.ts
```diff
@@ -0,0 +1,32 @@
+import { FORMAT_TO_MIME_TYPE } from './select-output-type.js'
+import type { ParsedUrlStringResults } from './parse-url-string.js'
+
+export interface IsAcceptExplicitOptions {
+
+  query?: ParsedUrlStringResults['query']
+  headers: Headers
+}
```
achingbrain, line 8:
```suggestion
export interface IsAcceptExplicitOptions {
  query?: ParsedUrlStringResults['query']
  headers: Headers
}
```

packages/verified-fetch/src/utils/get-resolved-accept-header.ts
```diff
@@ -0,0 +1,42 @@
+import { isExplicitAcceptHeader, isExplicitFormatQuery, isExplicitIpldAcceptRequest } from './is-accept-explicit.js'
+import { queryFormatToAcceptHeader } from './select-output-type.js'
+import type { ParsedUrlStringResults } from './parse-url-string.js'
+import type { ComponentLogger } from '@libp2p/interface'
+
+export interface ResolvedAcceptHeaderOptions {
+  query?: ParsedUrlStringResults['query']
+  headers?: RequestInit['headers']
+  logger?: ComponentLogger
```
achingbrain, line 9: `helia.logger` will always be set so I don't think this field needs to be optional.

achingbrain (changes requested):
LGTM, comments inline.

achingbrain (approved):
LGTM, comments inline.

PR #42, opened 2024-04-05
fix: use ipfs-unixfs-exporter directly

achingbrain:
`@helia/unixfs` pulls in the importer and other deps, we only need to stat and cat a DAG so we can just use the exporter for this.

## Change checklist


achingbrain:
Draft because the tests are failing

SgtPooki:
I like this much better than #32. not a priority, but we should investigate this path when we can

achingbrain:
This showed up a bug in the existing tests - the `should abort a request during unixfs.cat call` test waits on a promise before aborting the request - it aborts after `unixfs.stat` was called rather than `unixfs.cat`.

The desired behaviour is more complicated than it looks because the request is aborted during streaming the response rather than setting up the response as in the `should abort a request during unixfs.stat call` test.

Anyway, it's fixed now.

packages/verified-fetch/test/abort-handling.spec.ts
```diff
@@ -159,39 +125,137 @@ describe('abort-handling', function () {
...
+    expect(dnsLinkResolver.callCount).to.equal(0)
+
+    // not called because the blockstore has the content
+    expect(blockRetriever.retrieve.callCount).to.equal(0)
+
+    // the file root was loaded
+    expect(blockstoreGetSpy.getCalls().map(call => call.args[0].toString()))
+      .to.include(fileCid.toString())
+
+    // the leaf nodes were not have been loaded because the request was aborted
```
SgtPooki, line 180:
```suggestion
    // the leaf nodes were loaded because the request was aborted
```

packages/verified-fetch/test/abort-handling.spec.ts
```diff
@@ -159,39 +125,137 @@ describe('abort-handling', function () {
...
+
+    const fileRootGot = pDefer()
+    const blockstoreGetSpy = Sinon.stub(helia.blockstore, 'get')
+    blockstoreGetSpy.callsFake(async (cid, options) => {
+      if (cid.equals(fileCid)) {
+        fileRootGot.resolve()
+      }
+
+      return blockstoreGetSpy.wrappedMethod.call(helia.blockstore, cid, options)
+    })
```
SgtPooki, line 160: I was not familiar with `wrappedMethod`, nice.

SgtPooki (approved):
minor comment grammatical fix, otherwise looks good to me

packages/verified-fetch/test/abort-handling.spec.ts
achingbrain: Me neither, tbh.  It doesn't bind the right context so we have to use `.call` or `.apply` but it seems more concise than creating closures or whatever.

PR #45, opened 2024-04-08
fix: reduce dagPb and dagCbor handler complexity

### LINKED ISSUE #43: chore: reduce line complexity in `handleDagPb` and `handleDagCbor`
https://github.com/ipfs/helia-verified-fetch/pull/39#discussion_r1556124242

We need to break out some of the common functionality

SgtPooki:
## Title

fix: reduce dagPb and dagCbor handler complexity

## Description

fixes #43

- create and use `setIpfsRoots` function
- pull out walkPath logic into separate function

## Notes & open questions

I'm not 100% sure about how to handle the `isObjectNode === false` case.

## Change checklist


packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -259,7 +252,27 @@ export class VerifiedFetch {
...
+    if (pathDetails instanceof Response) {
+      return pathDetails
+    }
+    const ipfsRoots = pathDetails.ipfsRoots
+    if (isObjectNode(pathDetails.terminalElement)) {
+      terminalElement = pathDetails.terminalElement
+    } else {
+      // this should never happen, but if it does, we should log it and return notSupportedResponse
+      this.log.error('terminal element is not a dag-cbor node')
+      return notSupportedResponse(resource, 'Terminal element is not a dag-cbor node')
```
SgtPooki, line 272: not sure if this is the right thing to do here, or if we want to allow walking the path to fail and still query `helia.blockstore.get(cid, options)`

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -299,35 +312,23 @@ export class VerifiedFetch {
...
-      return badGatewayResponse(resource.toString(), 'Error walking path')
+    const pathDetails = await this.handlePathWalking({ cid, path, resource, options })
+    if (pathDetails instanceof Response) {
+      return pathDetails
     }
+    const ipfsRoots = pathDetails.ipfsRoots
+    const terminalElement = pathDetails.terminalElement
 
-    let resolvedCID = terminalElement?.cid ?? cid
+    let resolvedCID = terminalElement.cid
```
SgtPooki, line 331: do we want to keep the fallback to root CID? we shouldn't need to, but I could use a second set of eyes

SgtPooki (commented):
self review

packages/verified-fetch/src/types.ts
```diff
@@ -1,3 +1,32 @@
+import { type AbortOptions } from '@libp2p/interface'
+import { type CID } from 'multiformats/cid'
+import type { VerifiedFetchInit } from './index.js'
+
 export type RequestFormatShorthand = 'raw' | 'car' | 'tar' | 'ipns-record' | 'dag-json' | 'dag-cbor' | 'json' | 'cbor'
 
 export type SupportedBodyTypes = string | ArrayBuffer | Blob | ReadableStream<Uint8Array> | null
+
+export interface FetchHandlerFunctionArg {
```
SgtPooki, line 9: note: cacheKey was removed

packages/verified-fetch/src/utils/walk-path.ts
```diff
@@ -37,3 +40,23 @@ export async function walkPath (blockstore: ReadableStorage, path: string, optio
...
   return node.type === 'object'
 }
+
+/**
+ * Attempts to walk the path in the blockstore, returning ipfsRoots needed to resolve the path, and the terminal element.
+ * If the signal is aborted, the function will throw an AbortError
+ * If a terminal element is not found, a notFoundResponse is returned
+ * If another unknown error occurs, a badGatewayResponse is returned
+ */
+export async function handlePathWalking ({ cid, path, resource, options, blockstore, log }: Omit<FetchHandlerFunctionArg, 'session'> & { blockstore: Blockstore, log: Logger }): Promise<PathWalkerResponse | Response> {
```
SgtPooki, line 50: this was moved to `walk-path.ts` from `verified-fetch.ts` because that file is getting much too large

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -156,7 +126,8 @@ export class VerifiedFetch {
     this.log.trace('created VerifiedFetch instance')
   }
 
-  private getBlockstore (root: CID, key: string, useSession: boolean, options?: AbortOptions): Blockstore {
+  private getBlockstore (root: CID, resource: string | CID, useSession: boolean, options?: AbortOptions): Blockstore {
+    const key = resourceToSessionCacheKey(resource)
```
SgtPooki, line 130: instead of constructing the key in `fetch` function, construct in here and just use the resource that `FetchHandlerFunctionArg` already has.

SgtPooki (commented):
self review

PR #86, opened 2024-05-23
feat: dir index html listing

SgtPooki:
- feat: basic dir-index-html implementation
- fix: dir-index-html initial link clicking

## Title

feat: dir index html listing

## Description

Initial implementation of the directory index html listing.

Fixes https://github.com/ipfs/service-worker-gateway/issues/342

### Demo
https://github.com/ipfs/helia-verified-fetch/assets/1173416/bc4da706-e3de-41b0-99b4-c6c51f40a38b

## Notes & open questions

We don't want a templating engine, so code is going to be a bit verbose, but I plan to clean it up before removing this PR from draft.

This code is now based on the changes in https://github.com/ipfs/helia-verified-fetch/pull/169

NOTE: you can test dir-index-html listing by running the following from the root of the repo: `npm run build && node packages/gateway-conformance/dist/src/demo-server.js` and then visiting http://localhost:3441/ipfs/bafybeig6ka5mlwkl4subqhaiatalkcleo4jgnr3hqwvpmsqfca27cijp3i

## Change checklist


packages/verified-fetch/src/utils/dir-index-html.ts
lidel, line 1: ## Concern

Hm.. with all the CSS and SVG images that take hundreds of kilobytes, and reproducing look and feel for `boxo/gateway`, this feels like an overkill for this library, and more something that should live in separate package imported by upstream projects like https://github.com/ipfs/service-worker-gateway

Most of users of `verified-fetch` will use it programmatically to access files, they won't enumerate directories, and if they do, they would prefer deterministic JSON rather than a difficult to parse HTML. Dir listing is something specific to project built on top of this library.

Note that the same approach applies to DAG-CBOR – we are not generating user-friendly HTML  in verified-fetch.

## Some ideas

I vaguely remember discussing this in the past  (apologies, forgot with whom) and the idea was for  `verified-fetch` due its web dev nature to default (if no `Accept` is passed) to returning more useful JSON representation of UnixFS defined in  https://ipld.io/specs/codecs/dag-pb/spec/#logical-format (initial idea was to make it the same DAG-JSON as `ipfs dag get QmdC5Hav9zdn2iS75reafXBq1PH4EnqUmoxwoxkS5QtuME` so the DX/UX of switching from kubo RPC to verified-fetch is cushioned).

If  verified-fetch library returned dir listings as  JSON,  dir support does not introduce size/dependency overhead for users who only interact with files, and then SW/HTTP gateway could be converting JSON dir listings to user-friendly HTML, and we would not have to maintain any UI elements in this repository. 

If we want, a very basic unstyled HTML responses (`<ul>`) could be produced by verified-fetch, but only if request was done with `text/html` in  `Accept`.

lidel (changes requested):
Slight concern about generating styled HTML and bundling all the SVG images in this project.
Feels like something that should be done elsewhere – some ideas inline.

packages/verified-fetch/src/utils/dir-index-html.ts
SgtPooki: If we do return json, do we have tests for that in gateway conformance? otherwise, I think we should probably optimize for gateway conformance passing now (for feature parity) and improve this lib as gateway conformance has more of a targeted focus for use by verified-fetch

SgtPooki:
we need to create a wrapper module.. simile to how we do `@helia/*` that augments the functionality of verified-fetch

packages/verified-fetch/src/utils/dir-index-html.ts
lidel: Fair enough, but  if you see a test that makes no sense (overly specific to Kubo) then feel free to skip and fill issue in https://github.com/ipfs/gateway-conformance + ping me to expedite a new release.

btw: i think right now gateway-conformance expects HTML to be always supported, which is a bug imo.  Thats fine for now, but in the future we will make conformance tests send explicit `Accept` header with `text/html` (like we already do in for DAG-JSON)

SgtPooki:
Just so you know, this PR has been updated on top of https://github.com/ipfs/helia-verified-fetch/pull/169, and I added a test. Currently, the dir-index-html plugin is added in the verified-fetch package.json in the export map.

packages/verified-fetch/src/plugins/plugin-handle-dir-index-html.ts
```diff
@@ -0,0 +1,46 @@
...
+import { dirIndexHtml } from '../utils/dir-index-html.js'
+import { okResponse } from '../utils/responses.js'
+import { BasePlugin } from './plugin-base.js'
+import type { PluginContext, VerifiedFetchPluginFactory } from './types.js'
+
+export class DirIndexHtmlPlugin extends BasePlugin {
+  readonly codes = [dagPbCode]
+  canHandle (context: PluginContext): boolean {
+    const { cid, accept, pathDetails, directoryEntries } = context
+    this.log('checking if we can handle %c with accept %s', cid, accept)
```
SgtPooki, line 11:
```suggestion
```

packages/verified-fetch/src/plugins/plugin-handle-dir-index-html.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (directoryEntries?.length === 0) {
+      return false
+    }
+
+    return cid.code === dagPbCode
+  }
+
+  async handle (context: PluginContext): Promise<Response> {
+    const { resource, pathDetails, directoryEntries } = context
+    // const { getBlockstore } = this.pluginOptions
```
SgtPooki, line 28:
```suggestion
```

packages/verified-fetch/src/plugins/plugin-handle-dir-index-html.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (pathDetails?.terminalElement == null) {
+      throw new TypeError('Path details are required')
```
SgtPooki, line 31:
```suggestion
      throw new Error('Path details are required')
```

packages/verified-fetch/src/plugins/plugin-handle-dir-index-html.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    if (directoryEntries == null || directoryEntries?.length === 0) {
+      throw new TypeError('Directory entries are required')
```
SgtPooki, line 34:
```suggestion
      throw new Error('Directory entries are required')
```

packages/verified-fetch/src/plugins/plugin-handle-dag-pb.ts
```diff
@@ -94,9 +95,22 @@ export class DagPbPlugin extends BasePlugin {
...
+        try {
+          for await (const dirItem of fs.ls(dirCid, { signal: options?.signal, onProgress: options?.onProgress })) {
+            context.directoryEntries.push(dirItem)
+          }
+          // dir-index-html plugin or dir-index-json (future idea?) plugin should handle this
+          return null
+        } catch (e) {
+          log.error('error listing directory %c', dirCid, e)
+          return notSupportedResponse('Unable to get directory contents')
+        }
```
SgtPooki, line 113: this could be moved to dir-index-html plugin but I thought it made the most sense here.

SgtPooki (commented):
self review

PR #92, opened 2024-06-10
refactor: IPFS_NS_MAP from gateway-conformance v6

lidel:
This PR switches to v6 with changes from https://github.com/ipfs/gateway-conformance/pull/211

## Notes

- `IPFS_NS_MAP` handling refactored: No need to generate env from `dnslink.json`, a fixture in desired format is already provided, and we no longer mix subdomains with DNSLink tests, which also simplifies code

## Change checklist

### DIFF
--- a/.github/workflows/gateway-conformance.yml
+++ b/.github/workflows/gateway-conformance.yml
@@ -23,12 +23,12 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.21.x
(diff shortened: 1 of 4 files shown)


lidel:
@SgtPooki if this looks good, ok to merge? 

`v6` should cut the number of requests made to 1/3 (because `proxy-gateway` is now explicit test suite, and does not multiply every subdomain test anymore)

SgtPooki (approved):

PR #107, opened 2024-10-11
feat!: upgrade to helia 5

2color:
## Description 

- Upgrade Helia to v5 and js-libp2p to v2
- Fix breaking changes

## Change checklist


packages/verified-fetch/src/utils/get-tar-stream.ts
```diff
@@ -28,7 +27,7 @@ function toHeader (file: UnixFSEntry): Partial<TarEntryHeader> & { name: string
 
 function toTarImportCandidate (entry: UnixFSEntry): TarImportCandidate {
   if (!EXPORTABLE.includes(entry.type)) {
-    throw new CodeError('Not a UnixFS node', 'ERR_NOT_UNIXFS')
+    throw new Error('Not a UnixFS node')
```
achingbrain, line 30: You might want to use named errors for better DX here and elsewhere in this PR?

`@helia/unixfs` exports a `NotUnixFSError`, for example:

```ts
import { NotUnixFSError }  from '@helia/unixfs/errors'
```

achingbrain (changes requested):
LGTM, DX nit about throwing generic errors vs named errors.

We've lost the `.code` property so anyone making decisions on thrown errors needs another way to do it, `libp2p`/`helia` now use `.name` - here we're not using anything at all so the dev has to do it based on error messages which can be fragile.

2color:
> We've lost the .code property so anyone making decisions on thrown errors needs another way to do it, libp2p/helia now use .name - here we're not using anything at all so the dev has to do it based on error messages which can be fragile.

I've adapted to code to either reuse or define custom error types with the `name` property here.

2color:
It seems that somehow this PR has caused a regression in the Gateway Conformance. 

@SgtPooki Any pointers on debugging this?

achingbrain (approved):

achingbrain:
You may wish to change the PR title before merging.  `chore: ` will not cause a release but you've added `!` which means a breaking change so maybe you do want a release.  Suggest `deps!:` or `feat!:` or something.

achingbrain:
It doesn't look like the gateway conformance tests have passed for months: https://github.com/ipfs/helia-verified-fetch/actions/workflows/gateway-conformance.yml

Issue #106, opened 2024-09-20
docs: example how to enable both HTTP and libp2p at the same time


lidel:
> Extracting from discussions with @aschmahmann 

https://github.com/ipfs/helia-verified-fetch/tree/main/packages/verified-fetch#usage-with-customized-helia shows how to configure custom helia with HTTP-only(?) backend.  

iiuc `/wss` bitswap providers discovered over delegated routing API are dead and can't be used for bitswap due to lack of libp2p.

We should have an explicit example which enables both HTTP and bitswap-over-llibp2p backends at the same time, and make it a prominent one, then have HTTP-only after it.

lidel:
@2color @SgtPooki i feel it could also be me being out of loop for a while, is there a technical reason why we don't have http+libp2p mode, or is it just a matter of docs?

2color:
No technical reason. We could get one going.

SgtPooki:
FYI I also recently opened https://github.com/ipfs/helia/discussions/638

I don't think there's a technical reason, but I am in favor of adding an example that enables client-only-libp2p for doing bitswap without doing any dht-walking or other work. it should be a pretty simple config honestly, just https://github.com/ipfs/helia/blob/4f14996a9b976f2b60f4c8fe52a4fd1632420749/packages/helia/src/utils/libp2p-defaults.browser.ts#L34 without:

* bootstrap
* dht 
* autonat

2color:
I think we need to find a better way to allow users to customise Helia and libp2p when instantiating `VerifiedFetch`. By better I mean better than instantiating libp2p and Helia with custom configuration, as we do in the Service Worker. For most developers this approach is too complex and error prone.

Maybe it's also time for us to enable p2p retrieval by default in Verified Fetch. 

### How to do this today

```ts
const helia = await createHelia()
const verifiedFetch = await createVerifiedFetch(helia)
```

However, with this, you get hundreds of connections due to the bootstrap nodes and the DHT Client. Since Verified Fetch is concerned with retrieval 

### How can you do this without all the connections

```ts
// Create helia without connecting to any bootstrap peers
const helia = await createHelia({
  libp2p: {
    peerDiscovery: [],
  }
})
```

However, this will still instantiate the DHT service. Another problem is that this will also have all transports enabled, and we know some of them don't work well, e.g. WebTransport.

## Path forward

[...]

PR #130, opened 2024-11-05
feat: support p2p retrieval by default

2color:
## Description

- As discussed in https://github.com/ipfs/helia-verified-fetch/issues/106, this PR introduces new boolean options to enable transport specific direct retrieval.

## Notes & open questions

- Does this direction make sense?

## Change checklist


2color:
As discussed during Helia WG:

Moving forward suggestion we agreed on:
    - Only using helia-p2p
    - Remove the new options added by this PR

SgtPooki:
@2color I think we need to work on updating the deps across all packages and getting gateway conformance passing before this will pass. I'm focusing on [updating the deps](https://github.com/ipfs/helia-verified-fetch/pull/134), and once that is done, we should be able to pass this one.

packages/verified-fetch/src/index.ts
```diff
@@ -784,34 +792,75 @@ export interface VerifiedFetchInit extends RequestInit, ProgressOptions<BubbledP
...
+      libp2p,
       blockBrokers: [
         trustlessGateway({
           allowInsecure: init?.allowInsecure,
           allowLocal: init?.allowLocal
-        })
+        }),
+        bitswap()
       ],
+      dns: createDns(init?.dnsResolvers),
```
SgtPooki, line 848: FYI: setting `libp2pConfig.dns` was causing some failures in tests. do we need that? dns should be handled at the helia layer shouldn't it?

cc @2color @achingbrain

packages/verified-fetch/src/utils/get-peer-id-from-string.ts
```diff
@@ -0,0 +1,12 @@
+import { peerIdFromCID, peerIdFromString } from '@libp2p/peer-id'
+import { CID } from 'multiformats/cid'
+import type { PeerId } from '@libp2p/interface'
+
+export function getPeerIdFromString (peerIdString: string): PeerId {
```
SgtPooki, line 5: we had three places where we were doing libp2p peerId string parsing, so I pulled this into a single place.

SgtPooki (approved):
self review. 

@2color this is fully updated. mind taking a look at the code?

packages/verified-fetch/src/index.ts
```diff
@@ -784,34 +792,75 @@ export interface VerifiedFetchInit extends RequestInit, ProgressOptions<BubbledP
   allowInsecure?: boolean
 }
 
+type Libp2pServiceType = Pick<DefaultLibp2pServices, 'dcutr' | 'identify' | 'keychain' | 'ping' > & Record<`delegatedRouting${number}`, any>
```
2color, line 795: small nit. type is exported https://github.com/ipfs/helia-delegated-routing-v1-http-api/blob/1e60306ee391e9400dbc098fee5ebd4eb0bf7562/packages/client/src/index.ts#L162C18-L162C49 
```suggestion
type Libp2pServiceType = Pick<DefaultLibp2pServices, 'dcutr' | 'identify' | 'keychain' | 'ping' > & Record<`delegatedRouting${number}`, DelegatedRoutingV1HttpApiClient>
```

packages/verified-fetch/src/index.ts
```diff
@@ -784,34 +792,75 @@ export interface VerifiedFetchInit extends RequestInit, ProgressOptions<BubbledP
...
+
+    libp2pConfig.transports = [webSockets(), webRTCDirect()]
+
+    libp2pConfig.peerDiscovery = [] // disable default bootstrap peers
+    libp2pConfig.addresses = {} // disable default listen addresses
+
+    // We only need client/listen/fetch based services
+    const fetchOnlyServices = {
+      dcutr: libp2pConfig.services.dcutr,
+      dht: kadDHT({
```
2color, line 815: My intention was to disabling the DHT client. So this can be removed. 

Unless there's a strong reason to include, I think relying on delegated routing is an ok compromise.

achingbrain:
> chrome browser tests are failing in CI because playwright isn't installed

They're failing because https://github.com/microsoft/playwright shipped a breaking change - the fix is here: https://github.com/hugomrdias/playwright-test/pull/685

packages/verified-fetch/src/index.ts
2color: I think both Helia and libp2p rely on the DNS type. For libp2p it's mostly for resolving `dnsaddr` multiaddrs.

achingbrain:
The browser tests should run now but 74e27ac broke ts compilation so that'll need to be fixed first.

2color:
I realised that I didn't consider node.js in this PR. So some of the libp2p config modifications don't make sense.

packages/verified-fetch/src/utils/libp2p-types.ts
2color, line 1: I appreciate that you were able to fix this, but I'm worried that this is appeasing the TS compiler at the cost of the types actually helping us.

SgtPooki: The thing with the services type is that we don't really care since we don't ever touch them.. but I mostly just wanted to unblock the build. I am working on adding a test now.

sometimes, you have to coerce types, and I think the pros of coercing types outweigh the cons of adding a bunch of direct deps on all the libp2p things

packages/verified-fetch/src/utils/libp2p-types.ts
```diff
@@ -6,7 +6,9 @@ import type { Libp2pOptions } from 'libp2p'
 type BaseServiceMap = Record<string, unknown>
 type DelegatedRoutingServices = Record<`delegatedRouting${number}`, unknown | ((components?: unknown) => DelegatedRoutingV1HttpApiClient)>
 
-export interface Libp2pServices extends BaseServiceMap, DelegatedRoutingServices, Record<string, any> {
+// eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style
+export interface Libp2pServices extends BaseServiceMap, DelegatedRoutingServices {
+  [key: string]: any
```
achingbrain, line 11: This is a bit weird, `Record<string, any>` and `{ [key: string]: any }` are the same type.  How does the `Record` version break things?

SgtPooki:
```
[12:42:33] tsc [started]
src/index.ts(823,33): error TS2345: Argument of type 'VerifiedFetchLibp2p' is not assignable to parameter of type 'Libp2pOptions<{ [x: string]: ...; }> | undefined'.
  Type 'VerifiedFetchLibp2p' is not assignable to type 'Libp2pOptions<{ [x: string]: ...; }>'.
    Type 'VerifiedFetchLibp2p' is not assignable to type 'Libp2pInit<{ [x: string]: ...; }>'.
      Types of property 'services' are incompatible.
        Type 'Libp2pServices' is not assignable to type 'ServiceFactoryMap<{ [x: string]: ...; }>'.
          'string' index signatures are incompatible.
            Type 'unknown' is not assignable to type '(components: Components & { [x: string]: ...; }) => unknown'.
[12:42:34] tsc [failed]
[12:42:34] → Command failed with exit code 1: tsc
```

SgtPooki: We can do something like

```
export type Libp2pServices<T extends ServiceMap = ServiceMap> = {
  [Property in keyof T]: (components: any & T) => T[Property]
} & BaseServiceMap & DelegatedRoutingServices
```

¯\\_(ツ)\_/¯

SgtPooki:
There are so many fixes here that have nothing to do with libp2p or enabling p2p by default.. these changes should probably be pulled out into https://github.com/ipfs/helia-verified-fetch/pull/141

SgtPooki:
> There are so many fixes here that have nothing to do with libp2p or enabling p2p by default.. these changes should probably be pulled out into #141

This is done in [`70cde6f` (#141)](https://github.com/ipfs/helia-verified-fetch/pull/141/commits/70cde6ff1faf649122b9dd17a7057d77adb6959a)

2color:
I managed to somewhat simplify the types. Going to merge this so that we can publish the blog post that includes a runnable example. We can further iterate in followup PRs.

PR #140, opened 2024-11-19
chore: use grouping for dependabot

SgtPooki:
## Title

chore: use grouping for dependabot

## Description

Attempting to modify dependabot so that dependencies that should be grouped are.. this aims to prevent a bunch of single dep update PRs that require other deps to be updated at the same time.

## Change checklist


SgtPooki (commented):
self review. groupings might need polish, but this should help out a lot

PR #151, opened 2024-11-22
fix: add trace logging of libp2p config

2color:
## Description

While testing different bundlers/CDNs it seemed like the browser overrides we define in package.json aren't always respected. This gives us a way to check this quickly.

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/index.ts
+++ b/packages/verified-fetch/src/index.ts
@@ -840,7 +840,7 @@ export async function createVerifiedFetch (init?: Helia | CreateVerifiedFetchIni
       dns,
       routers
     })
-    init.logger.forComponent('helia:verified-fetch').trace('created verified-fetch with libp2p config:', JSON.stringify(libp2pConfig))
+    init.logger.forComponent('helia:verified-fetch').trace('created verified-fetch with libp2p config: %j', libp2pConfig)
   }
 
   const verifiedFetchInstance = new VerifiedFetchClass({ helia: init }, options)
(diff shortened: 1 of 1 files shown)


packages/verified-fetch/src/index.ts
```diff
@@ -840,6 +840,7 @@ export async function createVerifiedFetch (init?: Helia | CreateVerifiedFetchIni
       dns,
       routers
     })
+    init.logger.forComponent('helia:verified-fetch').trace('created verified-fetch with libp2p config:', JSON.stringify(libp2pConfig))
```
achingbrain, line 843: In general you want to avoid doing work like this for logging - the `JSON.stringify` will get invoked regardless of if logging is enabled or not.  It's not a big deal here since it's a one-off operation but it can degrade performance if you do it in a hot codepath.

Instead you can use `enabled`:

```js
import { enabled } from '@libp2p/logger'

if (enabled('helia:verified-fetch:trace')) {
  log.trace('expensive operation result: %s', JSON.stringify(libp2pConfig))
}
```

You should also include the `%s` token, otherwise it can mess up the formatting - in this case the `ms` diff since the last log will get printed before the `JSON.stringify` output.

achingbrain: You could also use the `%j` formatter here instead and skip the conditional:

```suggestion
    init.logger.forComponent('helia:verified-fetch').trace('created verified-fetch with libp2p config: %j', libp2pConfig)
```

SgtPooki (approved):

PR #156, opened 2024-12-02
feat: allow passing custom hashers to verified fetch

2color:
## Description

- **docs: add info about how blocks are retrieved**
- **feat: allow passing custom hashers**
- **docs: generate**
- **chore: add docs command**

## Notes & open questions

Passing custom hashers came up in a [Hacker News discussion](https://news.ycombinator.com/item?id=42263810). Since passing a custom Helia instance with libp2p requires a fair bit of domain knowledge, it seemed simple enough to bubble up to the Verified Fetch API surface. 

## Change checklist


packages/verified-fetch/src/index.ts
```diff
@@ -189,6 +194,24 @@
  *   }
  * })
  * ```
+ * ### Custom Hashers
+ *
+ * By default, `@helia/verified-fetch` supports `sha256`, `sha512`, and `identity` hashers.
```
SgtPooki, line 199: I would instead, or in addition, point to helia defaults.. that could help if this ever gets out of date

packages/verified-fetch/src/index.ts
```diff
@@ -657,6 +680,13 @@ export interface CreateVerifiedFetchInit {
    */
   dnsResolvers?: DNSResolver[] | DNSResolvers
 
+  /**
+   * By default sha256, sha512 and identity hashes are supported for
+   * retrieval operations. To retrieve blocks by CIDs using other hashes
+   * pass appropriate MultihashHashers here.
+   */
+  hashers?: MultihashHasher[]
```
SgtPooki, line 688: Can we "Pick" the type from Helia options so the types are tighter?

SgtPooki (approved):

SgtPooki:
Approval contingent on passing CI

2color:
Build problem:
- If I add `blake2b` as a dependency `dep-check` fails
- If I remove `blake2b` as a dependency, `lint` fails:

```
Error compiling example code block 10 in file README.md:
  ⨯ Unable to compile TypeScript:
      README.md → Code Block 10(2,28): error TS2307: Cannot find module '@multiformats/blake2/blake2b' or its corresponding type declarations.
```

I'll try adding a `// @ts-expect-error - blake2b256 is not a direct dependency`, but wondering if there's a better way to resolve this. @achingbrain

achingbrain:
I think it's because `dep-check` doesn't take code blocks in documentation into account when working out which dependencies are in use, which it shouldn't really as then we'd have dependabot updates (and potentially security notices) for dependencies that the shipping module code doesn't actually use.

I guess the choices are to either:

1. Mark the code snippet as JS instead of TS, this will mean we don't compile it so we miss any valid errors like changes in the API, boo
2. Disable this specific error in the `doc-check` command with something like https://github.com/ipfs/aegir/pull/1701

packages/verified-fetch/README.md
```diff
@@ -217,6 +225,27 @@ const fetch = await createVerifiedFetch({
...
+
+By default, `@helia/verified-fetch` supports `sha256`, `sha512`, and `identity` hashers.
+
+If you need to use a different hasher, you can provide a [custom `hasher` function](https://multiformats.github.io/js-multiformats/interfaces/hashes_interface.MultihashHasher.html) as an option to `createVerifiedFetch`.
+
+## Example - Passing a custom hashing function
+
+```typescript
+import { createVerifiedFetch } from '@helia/verified-fetch'
+// @ts-expect-error - blake2b256 is not a direct dependency
```
2color, line 238:
```suggestion
```

packages/verified-fetch/src/index.ts
```diff
@@ -189,6 +194,26 @@
...
+ *
+ * By default, `@helia/verified-fetch` supports `sha256`, `sha512`, and `identity` hashers.
+ *
+ * If you need to use a different hasher, you can provide a [custom `hasher` function](https://multiformats.github.io/js-multiformats/interfaces/hashes_interface.MultihashHasher.html) as an option to `createVerifiedFetch`.
+ *
+ * @example Passing a custom hashing function
+ *
+ * ```typescript
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ * // @ts-expect-error - blake2b256 is not a direct dependency
```
SgtPooki, line 207:
```suggestion
```

SgtPooki:
ship it!

PR #159, opened 2024-12-04
chore: fix docs and depcheck error

SgtPooki:
This is one way to resolve the issue mentioned by @2color in https://github.com/ipfs/helia-verified-fetch/pull/156#issuecomment-2515019000

it removes the need to add a //ts-expect-error in the example code generated on the README.md

### DIFF
--- a/packages/verified-fetch/.aegir.js
+++ b/packages/verified-fetch/.aegir.js
@@ -1,5 +1,10 @@
 /** @type {import('aegir').PartialOptions} */
 const options = {
+  dependencyCheck: {
+    ignore: [
+      '@multiformats/blake2'
+    ]
+  },
   build: {
     bundlesizeMax: '132KB'
   }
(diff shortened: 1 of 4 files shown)


achingbrain:
The problem with this approach, if we get into a habit of it, is that we will now receive dependabot updates for this extra dep, increasing maintenance friction, and potentially security notices if something nasty is discovered in the unused dep which causes noise.

SgtPooki:
This is not needed with the addition of https://github.com/ipfs/aegir/pull/1701

SgtPooki:
> The problem with this approach, if we get into a habit of it, is that we will now receive dependabot updates for this extra dep, increasing maintenance friction, and potentially security notices if something nasty is discovered in the unused dep which causes noise.

100%.. the PR you put up is the right long-term fix

2color:
Thanks for jumping in!

PR #163, opened 2024-12-09
feat: dynamic dag traversal

### LINKED ISSUE #52: Video streaming performance
Trying to load the [Big Buck Bunny CID](http://bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq.ipfs.localhost:8080/) via the http gateway takes about 15 seconds to start playing for me.  I've managed to get it down to 5-10 seconds with some bitswap optimisations but that's still a really long time.

What's happening is the DAG is wide so Helia gets loads of CID requests in quick succession and it takes time to resolve them.  This is great for fast file transfer but terrible for video streaming where we just want the first block as fast as possible.

Setting [blockReadConcurrency](https://github.com/ipfs/js-ipfs-unixfs/blob/main/packages/ipfs-unixfs-exporter/src/index.ts#L130) to `1` (e.g. read one block at a time) lowers playback start time to under a second.

To do this more properly we should probably sniff the content type of a file by reading the first block, then setting the block read concurrency accordingly (high if it's a file, low if it's a video or image (for progressive rendering?)).

This will be easier with #42 in.

SgtPooki:
- **refactor: move set-content-type function**
- **feat: enhanced dag traversal**
- **feat: improve enhanced-dag-traversal**

## Title

feat: dynamic dag traversal

## Description

This PR consolidates DAG traversal and streams when fetching dag-pb/unixfs. If the content type is an image or video, it attempts to return more quickly by doing a DFS traversal of the DAG and streaming the content as it is found. This is an enhancement to the existing DAG traversal and streaming mechanism.

A few callouts:

* We no longer need `getStreamFromAsyncIterable`
* firstChunk retrieval is eagerly done in `enhancedDagTraversal`
* all other chunks are lazily fetched in `enhancedDagTraversal`

Fixes #52

## Notes & open questions

It seems like we get a little performance boost with these changes. See https://github.com/ipfs/service-worker-gateway/pull/529 also hyperfine benchmarking below. Note that the first hyperfine run was querying public endpoints (without instantiating helia), and the second was querying a local endpoint

```console
> hyperfine --parameter-list branch main,52-video-streaming-performance --setup "git switch {branch} && npm run reset && npm i && npm run build && cd packages/verified-fetch" -i --runs 1000 "node test-time-to-first-byte.js"
Benchmark 1: node test-time-to-first-byte.js (branch = main)
  Time (mean ± σ):      24.0 ms ±   2.7 ms    [User: 17.8 ms, System: 3.5 ms]
  Range (min … max):    22.0 ms …  68.6 ms    1000 runs
 
  Warning: Ignoring non-zero exit code.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: node test-time-to-first-byte.js (branch = 52-video-streaming-performance)
  Time (mean ± σ):      23.4 ms ±   2.1 ms    [User: 17.7 ms, System: 3.4 ms]
  Range (min … max):    21.3 ms …  58.1 ms    1000 runs
 
  Warning: Ignoring non-zero exit code.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  node test-time-to-first-byte.js (branch = 52-video-streaming-performance) ran
    1.03 ± 0.15 times faster than node test-time-to-first-byte.js (branch = main)                                                                     [3m-6.517s]

> hyperfine --parameter-list branch main,52-video-streaming-performance --setup "git switch {branch} && npm run reset && npm i && npm run build" "cd packages/verified-fetch && node test-time-to-first-byte.js" --runs 5000
Benchmark 1: cd packages/verified-fetch && node test-time-to-first-byte.js (branch = main)
  Time (mean ± σ):     340.8 ms ±  15.3 ms    [User: 367.6 ms, System: 87.3 ms]
  Range (min … max):   320.7 ms … 638.8 ms    5000 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (638.8 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: cd packages/verified-fetch && node test-time-to-first-byte.js (branch = 52-video-streaming-performance)
  Time (mean ± σ):     341.5 ms ±  11.5 ms    [User: 368.4 ms, System: 87.4 ms]
  Range (min … max):   322.6 ms … 599.5 ms    5000 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (561.6 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Summary
  cd packages/verified-fetch && node test-time-to-first-byte.js (branch = main) ran
    1.00 ± 0.06 times faster than cd packages/verified-fetch && node test-time-to-first-byte.js (branch = 52-video-streaming-performance)           [60m-28.332s]
```

The contents of `test-time-to-first-byte.js` is:

```js
import { trustlessGateway } from '@helia/block-brokers'
import { createHeliaHTTP } from '@helia/http'
import { httpGatewayRouting } from '@helia/routers'
import browserReadableStreamToIt from 'browser-readablestream-to-it'
import first from 'it-first'
import { createVerifiedFetch } from './dist/src/index.js'

const controller = new AbortController()
const signal = controller.signal
const start = performance.now()
const helia = await createHeliaHTTP({
  blockBrokers: [trustlessGateway({ allowLocal: true })],
  routers: [httpGatewayRouting({ gateways: ['http://127.0.0.1:8080'] })]
})
const verifiedFetch = await createVerifiedFetch(helia)
const response = await verifiedFetch('bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq', { signal })

const end = performance.now()
const timeToResponse = end - start
if (response.body == null) {
  throw new Error('response.body is null')
}
const startByte = performance.now()
await first(browserReadableStreamToIt(response.body))
const endByte = performance.now()
const timeToFirstByte = endByte - startByte
// expect(timeToFirstByte).to.be.lessThan(1000)
// eslint-disable-next-line no-console
console.log('TTR: %s, TTFB: %s', timeToResponse, timeToFirstByte)

await verifiedFetch.stop()
controller.abort()
```

Things still to do:

- [x] Test with sw-gateway - https://github.com/ipfs/service-worker-gateway/pull/529
- [ ] Test with helia-http-gateway
- [ ] Fully remove `getStreamFromAsyncIterable` and `getStreamFromAsyncIterable` tests
- [ ] Add tests for `enhancedDagTraversal`
- [ ] Unskip & fix abort-handling tests

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
+++ b/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
@@ -26,8 +26,9 @@ export async function enhancedDagTraversal ({
   const peekableIter = peekable(entry.content({
     signal,
     onProgress,
-    offset,
-    length: 50,
+    offset: 0,
+    // https://pkg.go.dev/net/http#DetectContentType reads the first 512 bytes
+    length: 512,
     blockReadConcurrency: 1
   }))
 
--- a/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
+++ b/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
@@ -27,7 +27,7 @@ export async function enhancedDagTraversal ({
     signal,
     onProgress,
     offset,
-    length,
+    length: 50,
     blockReadConcurrency: 1
   }))
 
--- a/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
+++ b/packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
@@ -1,18 +1,10 @@
 import { type ComponentLogger } from '@libp2p/interface'
-import { type ReadableStorage, exporter, type ExporterOptions, type UnixFSEntry } from 'ipfs-unixfs-exporter'
-// import first from 'it-first'
+import { type ExporterOptions, type UnixFSEntry } from 'ipfs-unixfs-exporter'
 import peekable from 'it-peekable'
 import toBrowserReadableStream from 'it-to-browser-readablestream'
-import { type CID } from 'multiformats/cid'
-import { type ContentTypeParser } from '../types.js'
-import { getContentType } from './get-content-type.js'
 
 export interface EnhancedDagTraversalOptions extends ExporterOptions {
-  blockstore: ReadableStorage
-  cidOrPath: string | CID
   logger: ComponentLogger
-  path: string
-  contentTypeParser?: ContentTypeParser
   entry: UnixFSEntry
 }
 
@@ -22,39 +14,33 @@ export interface EnhancedDagTraversalResponse {
 }
 
 export async function enhancedDagTraversal ({
-  blockstore,
   signal,
   onProgress,
-  cidOrPath,
   offset,
   length,
-  path,
   logger,
-  contentTypeParser,
   entry
 }: EnhancedDagTraversalOptions): Promise<EnhancedDagTraversalResponse> {
   const log = logger.forComponent('helia:verified-fetch:enhanced-dag-traversal')
 
-  const dfsIter = entry.content({
+  const peekableIter = peekable(entry.content({
     signal,
     onProgress,
     offset,
     length,
     blockReadConcurrency: 1
-  })
+  }))
 
   let firstChunk: Uint8Array = new Uint8Array()
   let error: Error
   try {
-    // Fetch the first chunk eagerly
-    const peekableIter = peekable(dfsIter)
+    // Fetch the first chunk eagerly to determine the content type
     const firstPeek = await peekableIter.peek()
     if (firstPeek.done === true) {
       throw new Error('No content found')
     }
     firstChunk = firstPeek.value
     peekableIter.push(firstChunk)
-    // firstChunk = await first(dfsIter)
   } catch (err: any) {
     if (signal?.aborted === true) {
       error = err
@@ -71,13 +57,6 @@ export async function enhancedDagTraversal ({
           throw error
         }
 
-        // // Determine content type based on the first chunk
-        // const contentType = await getContentType({ bytes: firstChunk, contentTypeParser, path, log })
-
-        // const isImageOrVideo = contentType.startsWith('video/') || contentType.startsWith('image/')
-        // log.trace('Content type determined: %s', contentType)
-
-        // continue with the BFS iterator
         for await (const chunk of entry.content({
           signal,
           onProgress,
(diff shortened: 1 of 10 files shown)


packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -375,22 +375,16 @@ export class VerifiedFetch {
...
-
-      const asyncIter = entry.content({
+      const { firstChunk, stream } = await enhancedDagTraversal({
+        blockstore: this.helia.blockstore,
         signal: options?.signal,
         onProgress: options?.onProgress,
+        cidOrPath: resolvedCID,
         offset,
-        length
-      })
```
SgtPooki, line 388: we can pass the `terminalElement` here because that comes from `walkPath` from ipfs-unixfs-exporter which is an Entry, the same as return type from `exporter`

packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
```diff
@@ -0,0 +1,104 @@
...
+  logger,
+  contentTypeParser
+}: EnhancedDagTraversalOptions): Promise<EnhancedDagTraversalResponse> {
+  const log = logger.forComponent('helia:verified-fetch:enhanced-dag-traversal')
+
+  const dfsEntry = await exporter(cidOrPath, blockstore, {
+    signal,
+    onProgress,
+    blockReadConcurrency: 1
+  })
```
SgtPooki, line 40: we should receive `entry` and not need to call `exporter` here at all

SgtPooki (commented):
some comments after chatting with Alex in Helia WG

packages/verified-fetch/src/utils/enhanced-dag-traversal.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const dfsIter = dfsEntry.content({
+    signal,
+    onProgress,
+    offset,
+    length
+  })
+
```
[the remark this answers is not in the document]
SgtPooki: updated this to follow the same pattern as https://pkg.go.dev/net/http#DetectContentType for detecting the content type

SgtPooki:
Waiting for review here since this is a significant change.. maybe i'm missing something (and I haven't looked since last year.. har har) but I did not see a huge improvement.

SgtPooki:
~~update benchmarks and tests to use ipfsd-ctl to start up the kubo node, so its a new gateway for each request that doesn't have the blocks cached in the response.~~

instead simulate latency by delaying for at least a second to test in the worst case. this can be done by wrapping the helia blockstore with a custom one where we delay in the get method.

PR #164, opened 2024-12-13
feat: add server timing to responses

SgtPooki:
## Title

feat: add server timing to responses

## Description

This PR adds server timing headers to responses by wrapping all responses returned by `fetch` in a `Response` object that includes the server timing headers.

The headers are generated by wrapping async functions with a `serverTiming` utility function that records the time taken to execute the function, and then adding those headers to an array on the VerifiedFetch class.

## Notes & open questions

We need to work out exactly what events we need to time, and I'd also like to clean up the `serverTiming` utility function usage so that it's easier to add new timings without having to handle the error and success cases manually for each one.

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/verified-fetch.ts
+++ b/packages/verified-fetch/src/verified-fetch.ts
@@ -32,6 +32,7 @@ import { resourceToSessionCacheKey } from './utils/resource-to-cache-key.js'
 import { setCacheControlHeader, setIpfsRoots } from './utils/response-headers.js'
 import { badRequestResponse, movedPermanentlyResponse, notAcceptableResponse, notSupportedResponse, okResponse, badRangeResponse, okRangeResponse, badGatewayResponse, notFoundResponse } from './utils/responses.js'
 import { selectOutputType } from './utils/select-output-type.js'
+import { serverTiming } from './utils/server-timing.js'
 import { setContentType } from './utils/set-content-type.js'
 import { handlePathWalking, isObjectNode } from './utils/walk-path.js'
 import type { CIDDetail, ContentTypeParser, CreateVerifiedFetchOptions, Resource, ResourceDetail, VerifiedFetchInit as VerifiedFetchOptions } from './index.js'
@@ -111,6 +112,7 @@ export class VerifiedFetch {
   private readonly log: Logger
   private readonly contentTypeParser: ContentTypeParser | undefined
   private readonly blockstoreSessions: LRUCache<string, SessionBlockstore>
+  private serverTimingHeaders: string[] = []
 
   constructor ({ helia, ipns }: VerifiedFetchComponents, init?: CreateVerifiedFetchOptions) {
     this.helia = helia
@@ -254,7 +256,11 @@ export class VerifiedFetch {
     const blockstore = this.getBlockstore(cid, resource, session, options)
 
     // need to walk path, if it exists, to get the terminal element
-    const pathDetails = await handlePathWalking({ cid, path, resource, options, blockstore, log: this.log })
+    const { error, result: pathDetails, header: pathWalkingHeader } = await serverTiming('path-walking', '', async () => handlePathWalking({ cid, path, resource, options, blockstore, log: this.log }))
+    if (error != null) {
+      throw error
+    }
+    this.serverTimingHeaders.push(pathWalkingHeader)
     if (pathDetails instanceof Response) {
       return pathDetails
     }
@@ -316,7 +322,12 @@ export class VerifiedFetch {
     let redirected = false
     const byteRangeContext = new ByteRangeContext(this.helia.logger, options?.headers)
     const blockstore = this.getBlockstore(cid, resource, session, options)
-    const pathDetails = await handlePathWalking({ cid, path, resource, options, blockstore, log: this.log })
+    const { error, result: pathDetails, header: pathWalkingHeader } = await serverTiming('path-walking', '', async () => handlePathWalking({ cid, path, resource, options, blockstore, log: this.log }))
+    if (error != null) {
+      throw error
+    }
+    this.serverTimingHeaders.push(pathWalkingHeader)
+
     if (pathDetails instanceof Response) {
       return pathDetails
     }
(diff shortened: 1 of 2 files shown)


SgtPooki:
with https://github.com/ipfs/service-worker-gateway/pull/531

we get:

image

SgtPooki (commented):
self review

SgtPooki (commented):
self review of added test

2color (commented):
This is pretty neat. Implementation looks good to me.

What do you think about making this optional per request? I can imagine that in node.js this isn't very useful and just makes the response object larger. 

> We need to work out exactly what events we need to time

At the very least we'd want: 
- IPNS/DNSLink resolution (if it's an ipns url) 
- content routing, i.e. finding providers.

SgtPooki:
@2color added the option to include server timing headers, defaulting to false on construction but overridable per request. Also added docs and updated readme

packages/verified-fetch/src/index.ts
```diff
@@ -604,6 +604,14 @@
  * * https://specs.ipfs.tech/http-gateways/trustless-gateway/#response-headers
  * * https://specs.ipfs.tech/http-gateways/subdomain-gateway/#response-headers
  *
+ * #### Server Timing headers
+ *
+ * By default, we do not include Server Timing headers in responses. If you want to include them, you can pass an
```
2color, line 609:
```suggestion
 * By default, we do not include [Server Timing](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Server_timing) headers in responses. If you want to include them, you can pass an
```

packages/verified-fetch/src/index.ts
```diff
@@ -750,6 +758,14 @@ export interface CreateVerifiedFetchOptions {
...
    */
   sessionTTLms?: number
+
+  /**
+   * Whether to include server-timing headers in responses. This option can be overridden on a per-request basis.
+   *
+   * @default false
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing
+   */
+  includeServerTiming?: boolean
```
2color, line 768: Since it's typed as a boolean, any thoughts on making this a bit shorter?
```suggestion
  serverTiming?: boolean
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -452,6 +460,26 @@ export class VerifiedFetch {
     [identity.code]: this.handleRaw
   }
 
+  private async handleServerTiming<T> (name: string, description: string, fn: () => Promise<T>): Promise<T> {
```
2color, line 463: Can we check `this.includeServerTiming`  here to encapsulate most of the logic here and avoid the measuring and assigning of `this.serverTimingHeaders`? 

That would also allow simplifying `responseWithServerTiming`  and renaming to something a bit more generic, e.g. `augmentResponse` or `handleResponse`

SgtPooki: yea let me peek at this

packages/verified-fetch/src/index.ts
SgtPooki: I don't have a strong preference, but I prefer a more explicit name since `serverTiming` doesn't communicate its purpose. Also, verb-prefixed names for booleans are more idiomatic. What do you think about `withServerTiming` instead?

Option names should communicate their intent and effect rather than aim for brevity. More explicit names lead to a better dev understanding, with and without reading the docs. Also, the options are only passed once.

2color: Fair enough! I agree. 

`withServerTiming` works too and is shorter.

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: so, it just so happened that switching the server timing handling to handleServerTiming required me to pipe things through a lot more, so I'm moving forward with this and adding the "IPNS/DNSLink resolution (if it's an ipns url)" you previously mentioned.

I didn't originally like it in `handleServerTiming` because it required plumbing the request-level "withServerTiming" into more places, but it makes more sense for multiple reasons. 

it also prevents the extra work of wrapping the timed functions.. so it's a win. thanks for the push

SgtPooki:
@2color all requested changes are in, but the code is a little bit sloppier now because of how parse-resource & parse-url-string are currently written.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -207,7 +225,19 @@ export async function parseUrlString ({ urlString, ipns, logger }: ParseUrlStrin
...
-          resolveResult = await ipns.resolveDNSLink(decodedDnsLinkLabel, options)
+          // eslint-disable-next-line max-depth
+          if (withServerTiming) {
+            const resolveResultWithServerTiming = await serverTiming('ipns.resolveDNSLink', `Resolve DNSLink ${decodedDnsLinkLabel}`, ipns.resolveDNSLink.bind(ipns, decodedDnsLinkLabel, options))
+            serverTimings.push(resolveResultWithServerTiming)
+            // eslint-disable-next-line max-depth
+            if (resolveResultWithServerTiming.error != null) {
+              throw resolveResultWithServerTiming.error
+            }
+            resolveResult = resolveResultWithServerTiming.result ?? undefined
```
2color, line 236:
```suggestion
            resolveResult = resolveResultWithServerTiming.result
```

2color (commented):
LGTM. I still think it would be really useful to add a duration for finding providers from each router/block-broker. 

Since that would probably require us to add progress hooks to the helia routers, we can handle this in a follow up PR. 

Not directly related, but more insight into routing would help surface more nuanced errors: no providers vs. failure to connect to providers, to the user.

SgtPooki:
CI seems to be failing when attempting to install Kubo on windows.. i'm looking into it and will try clearing package-lock and updating Kubo

SgtPooki:
> CI seems to be failing when attempting to install Kubo on windows.. i'm looking into it and will try clearing package-lock and updating Kubo

Nikolai is looking into the collab cluster

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: this was needed for typescript and then I updated some code.. just checked if we can remove the null coalescing to undefined and it seems like we can.. just pushed up the change

PR #186, opened 2025-02-28
chore: do not start kubo node if KUBO_GATEWAY is set

SgtPooki:
## Change checklist


SgtPooki (commented):
self review

PR #188, opened 2025-03-04
fix: limit size of firstChunk

SgtPooki:
## Title

fix: limit size of firstChunk to 8kB

## Description

Addresses an issue brought up by @lidel where the first chunk of content was being read in its entirety, which could lead to memory exhaustion for large files. This PR limits the size of the first chunk to 8kB.

We only use this first chunk for content-type recognition.

## Change checklist

### DIFF
--- a/packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
+++ b/packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
@@ -2,6 +2,7 @@ import { defaultLogger } from '@libp2p/logger'
 import { expect } from 'aegir/chai'
 import sinon from 'sinon'
 import { getStreamFromAsyncIterable } from '../src/utils/get-stream-from-async-iterable.js'
+import { equals } from 'uint8arrays/equals'
 
 describe('getStreamFromAsyncIterable', () => {
   let onProgressSpy: sinon.SinonSpy
@@ -19,7 +20,7 @@ describe('getStreamFromAsyncIterable', () => {
     const chunks = new TextEncoder().encode('Hello, world!')
     const iterator = (async function * () { yield chunks })()
     const { firstChunk, stream } = await getStreamFromAsyncIterable(iterator, 'test.txt', defaultLogger(), { onProgress: onProgressSpy })
-    expect(firstChunk).to.equal(chunks)
+    expect(equals(firstChunk, chunks)).to.equal(true)
     const reader = stream.getReader()
     const { value } = await reader.read()
     expect(onProgressSpy.callCount).to.equal(1)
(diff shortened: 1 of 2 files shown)


packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
```diff
@@ -19,7 +20,7 @@ describe('getStreamFromAsyncIterable', () => {
     const chunks = new TextEncoder().encode('Hello, world!')
     const iterator = (async function * () { yield chunks })()
     const { firstChunk, stream } = await getStreamFromAsyncIterable(iterator, 'test.txt', defaultLogger(), { onProgress: onProgressSpy })
-    expect(firstChunk).to.equal(chunks)
+    expect(equals(firstChunk, chunks)).to.equal(true)
```
achingbrain, line 23: You can use a simpler assertion:

```suggestion
    expect(firstChunk).to.equalBytes(chunks)
```

achingbrain:
Each chunk the exporter yields corresponds to a block.  These are normally 256KiB or 1MiB but depend on the DAG layout.

SgtPooki:
> Each chunk the exporter yields corresponds directly to a block. These are normally 256KiB or 1MiB but it depends on the DAG layout.

I believe the concern here was that with HTTP retrieval in kubo/boxo, the blocks could be larger than we're expecting.. but helia squashes that in the blockstore, so we would still have our own block sizes, making provider chunk sizes irrelevant, yes?

achingbrain:
> with HTTP retrieval in kubo/boxo, the blocks could be larger than we're expecting..

As in, you could get a CAR that contains 1TB blocks?  Or something else?

lidel:
A CAR or a single Block.

The underlying issue is about enforcing some default max chunk size at retrieval layer. 

Passing too much to content-type sniffer (`file-type`) is not the best, there is no point in passing more than 8KiB, `file-type` default for inspecting streams has `sampleSize=4100`, but that is a side-quest nit, not the main problem.

The main problem is there is nothing blocking someone from announcing a CID that is a single big raw block (let's say 4GiB or 350GiB) and then tricking client into attempting to retrieve it.

If verified-fetch is backed by both HTTP and libp2p retrieval:
- `@helia/bitswap` client has block size limits enforced via `maxIncomingMessageSize` and `maxOutgoingMessageSize` which are [capped at 4MiB here](https://github.com/ipfs/helia/blob/375796aaead36111c4d663b061bf0edfe01c62ca/packages/bitswap/src/constants.ts#L12), making sure there is no surface for tricking client into slurping gigabytes of useless data.
- `@helia/http` client – afaik, maybe I missed something? – does not have ANY limit , so it will make HTTP GET for `https://example.com/ipfs/cid?format=raw` and slurp gigabytes. Trick client to attempt to fetch a few of such CIDs  and you killed http retrieval, because all free slots are taken and busy with meaningless work. 
   - :point_right:  I think the fix for now is to ensure `@helia/http` client applies same ballpark block size limit as `@helia/bitswap` client (or make it explicit cap at 2MiB per block).

achingbrain:
Fixing `@helia/http` to prevent endless downloads sounds better approach - once the block is in the blockstore it's too late.

It should check the content-length header before continuing the download of a block (if present) but it should also count the downloaded bytes in case the header is a lie.

SgtPooki:
Closing this, we should discuss any further at https://github.com/ipfs/helia/issues/755

Issue #165, opened 2024-12-23
feat: add Content-Length Header when possible


lhns:
Is it possible to know the object size before downloading?
This would allow showing a detailed progress bar for example.
If so it could be added as the Content-Length header to the response object.

SgtPooki:
There are some cases in which we may know the content length, and we should add the header in those cases IMHO, but it's not currently specified in the [trustless gateway specification](https://specs.ipfs.tech/http-gateways/trustless-gateway/), which we're targeting first with `@helia/verified-fetch`. 

There are still some unknowns and things to clarify here, but we might be able to add a content-length header in some cases. 

**Relevant specs & issues:**

* https://specs.ipfs.tech/http-gateways/path-gateway/#content-length-response-header
* https://github.com/ipfs/specs/issues/461
* https://specs.ipfs.tech/http-gateways/trustless-gateway/ (no spec for content length)

---

For now, you may be able to use [HTTP Range request headers](https://specs.ipfs.tech/http-gateways/path-gateway/#range-request-header) to figure out more about the size of the content you're fetching.  

[E.g](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range#syntax). `Range: bytes=0-` 

And then you would get back a content-range header in the response with the size (if we can determine it): `Content-Range: <unit> <range-start>-<range-end>/<size>`

The logic for byte range handling is in https://github.com/ipfs/helia-verified-fetch/blob/main/packages/verified-fetch/src/utils/byte-range-context.ts#L9-L28

lhns:
fixed by #189

PR #189, opened 2025-03-04
fix: add content-length header when possible

SgtPooki:
## Title

fix: add content-length header when we can

## Description

Adds a `Content-Length` header to the request when we can determine the length of the body.

This change also makes sure that all plugins support byte-ranges now.

Related https://github.com/ipfs/helia-verified-fetch/issues/165

## Change checklist


packages/verified-fetch/src/utils/get-stream-from-async-iterable.ts
```diff
@@ -46,6 +46,6 @@ export async function getStreamFromAsyncIterable (iterator: AsyncIterable<Uint8A
 
   return {
     stream,
-    firstChunk
+    firstChunk: firstChunk.slice(0, 8192)
```
SgtPooki, line 49: remove this and changes to get-stream-from-async-iterable.spec.ts

SgtPooki (commented):
self review

PR #192, opened 2025-03-06
fix: migrate from forked file-type to original

SgtPooki:
## Title

fix: migrate from forked file-type to original 

## Description

Tests currently failing locally

Related https://github.com/ipfs/helia-verified-fetch/issues/191#issuecomment-2704107574

## Notes & open questions

seems to make no difference in size.. this is expected and ideal.

### Before change:

image

### After change:

image

## Change checklist


SgtPooki (commented):
self review

SgtPooki:
This is not passing locally on macos for me.. even after pulling down a fresh clone for this branch. So i am not sure we can switch back to file-type yet.

SgtPooki:
> I ran the tests with this PR locally it passed (for node.js). Which test failed for you @SgtPooki? Was it in a browser runtime?

It was gateway conformance.. 

```shell
@helia/verified-fetch-gateway-conformance:          TestDagPbConversion has a success rate of at least 35.38%:
@helia/verified-fetch-gateway-conformance:      Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/sgtpooki/code/work/ipshipyard/ipfs/helia-verified-fetch/packages/gateway-conformance/dist/src/conformance.spec.js)
```

i'm thinking I need a restart.. it's failing on other runs too.. not sure what's going on. I think since it passed in CI and passed for you we can proceed.. I will update this PR once https://github.com/ipfs/helia-verified-fetch/pull/193 is in (or I can update file-type in that pr)

SgtPooki (commented):
self review

PR #193, opened 2025-03-06
feat: provide default content-type-parser

### LINKED ISSUE #191: deduplicate content-type-parser.ts
We seem to have the same code related to content type parsersing in multiple places:

- https://github.com/search?q=org%3Aipfs+application%2Fvnd.ms-powerpoint&type=code (!)
  - https://github.com/ipfs/service-worker-gateway/blob/957a9506c6c89b069d27ca4f6986e9ee1c6ef4a3/src/lib/content-type-parser.ts
  - https://github.com/ipfs/helia-http-gateway/blob/8c568ffb940846ef73e1d8b31f793e9d42d3c319/src/content-type-parser.ts
  - https://github.com/ipfs/helia-verified-fetch/blob/c36970d615bb07113b8971c3ec88e36f9b30f1f6/packages/gateway-conformance/src/fixtures/content-type-parser.ts

There should be a single source of truth, having unique copy used only in conformance tests reduces trust in having meaningful `content-type` tests anywhere

SgtPooki:
## Title

feat: provide default content-type-parser

## Description

Verified-fetch will now come with a default content-type-parser that will be used if no parser is provided. This parser will be able to parse JSON, text, and binary data.

Fixes #191

## Before Change

image

### After change

image

## Change checklist


2color (dismissed):

packages/verified-fetch/test/content-type-parser.spec.ts
```diff
@@ -28,24 +28,29 @@ describe('content-type-parser', () => {
...
   })
 
-  it('sets default content type if contentTypeParser is not passed', async () => {
-    verifiedFetch = new VerifiedFetch({
-      helia
-    })
-    const resp = await verifiedFetch.fetch(cid)
-    expect(resp.headers.get('content-type')).to.equal('application/octet-stream')
-  })
+  // it('sets default content type if contentTypeParser is not passed', async () => {
```
achingbrain, line 47: Delete test instead of commenting out?

achingbrain (commented):
LGTM

packages/verified-fetch/test/content-type-parser.spec.ts
SgtPooki: leftover debugging.. let me look again

PR #195, opened 2025-03-07
fix: go bin path

2color:
## Title

Fix the way the go bin path is constructed if GOPATH is set. Make sure we also add `bin` to the path, because binaries live in the `$GOPATH/bin`.

See https://go.dev/wiki/GOPATH


SgtPooki (approved):
lgtm.. conformance tests still run for me (no GOPATH set)

PR #197, opened 2025-03-11
fix: etag is path aware

### LINKED ISSUE #196: bug: etag not set correctly for paths
when requesting http://bafybeifiugn3zaj67dh6zqlgqpnzkbykbec3iur6wpb2zaz44ekcazyew4.ipfs.localhost:8345/image.jpeg, we should not be setting etag to `"bafybeifiugn3zaj67dh6zqlgqpnzkbykbec3iur6wpb2zaz44ekcazyew4"`, but we currently are.

SgtPooki:
## Title

fix: etag is path aware

## Description

Update plugin types & make work with `handleFinaleResponse`, clean up `fetch` impl internals, add get-e-tag test to verify path-awareness.

Fixes #196

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/verified-fetch.ts
+++ b/packages/verified-fetch/src/verified-fetch.ts
@@ -26,7 +26,6 @@ import { selectOutputType } from './utils/select-output-type.js'
 import { serverTiming } from './utils/server-timing.js'
 import type { CIDDetail, ContentTypeParser, CreateVerifiedFetchOptions, Resource, ResourceDetail, VerifiedFetchInit as VerifiedFetchOptions } from './index.js'
 import type { VerifiedFetchPlugin, PluginContext, PluginOptions } from './plugins/types.js'
-import type { RequestFormatShorthand } from './types.js'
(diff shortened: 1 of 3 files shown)


packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -199,7 +188,7 @@ export class VerifiedFetch {
     }
 
     if (cid != null && response.headers.get('etag') == null) {
-      response.headers.set('etag', getETag({ cid, reqFormat, weak: false }))
+      response.headers.set('etag', getETag({ cid: pathDetails?.terminalElement.cid ?? cid, reqFormat, weak: false }))
```
SgtPooki, line 191: this is the actual etag fix

SgtPooki (commented):
self review

packages/verified-fetch/src/verified-fetch.ts
Mkirby8867: This is confusing

PR #207, opened 2025-03-26
feat: support multiple byte-ranges

SgtPooki:
## Title

feat: support multiple byte-ranges

## Description

Add support for multiple byte ranges. Also updated byte-range-context's `setBody` method to allow for a function to be passed in. This function is called with each byte-range and creates a stream for the entire response body.

Related https://github.com/ipfs/helia-verified-fetch/pull/189

## Change checklist

### DIFF
--- a/packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
+++ b/packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
@@ -2,6 +2,7 @@ import { defaultLogger } from '@libp2p/logger'
 import { expect } from 'aegir/chai'
 import sinon from 'sinon'
 import { getStreamFromAsyncIterable } from '../src/utils/get-stream-from-async-iterable.js'
+import { equals } from 'uint8arrays/equals'
 
 describe('getStreamFromAsyncIterable', () => {
   let onProgressSpy: sinon.SinonSpy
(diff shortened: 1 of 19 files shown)


packages/verified-fetch/test/range-requests.spec.ts
```diff
@@ -129,13 +129,24 @@ describe('range requests', () => {
...
       })
 
-      it('should return 416 Range Not Satisfiable when passed multiple ranges', async () => {
-        await assertFailingRange(verifiedFetch.fetch(cid, {
-          headers: {
-            Range: 'bytes=0-2,3-5'
-          }
-        }))
-      })
+      // it('should return valid response when passed multiple ranges', async () => {
```
SgtPooki: re-enable and fix

packages/verified-fetch/test/get-stream-from-async-iterable.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { defaultLogger } from '@libp2p/logger'
 import { expect } from 'aegir/chai'
 import sinon from 'sinon'
+import { equals } from 'uint8arrays/equals'
```
SgtPooki: revert these changes

SgtPooki (commented):
self review

SgtPooki (commented):
self review.. a lot of things in here now. I know there was some concern about maintaining code for supporting multiple byte-ranges, but this essentially fixed #250 and gave us a more robust library. the range support is the same for single and multiple ranges now too.

PR #214, opened 2025-04-18
fix: aborted signal is handled when walking path

SgtPooki:
## Title

fix: aborted signal is handled when walking path

## Description

Note, this fix was manually confirmed as fixing the issue mentioned in https://github.com/ipfs/helia/issues/780 in the service-worker-gateway.

Related https://github.com/ipfs/service-worker-gateway/pull/681
Related https://github.com/ipfs/helia/issues/780

## Notes & open questions

Note the comments at https://github.com/ipfs/service-worker-gateway/pull/681#issuecomment-2816001952. There seems to be a bug in the abort handling with either libp2pRouting, bitswap sessions, and/or `ipfs-unixfs-exporter`, but I cannot reproduce it.. see https://github.com/SgtPooki/repro-helia-libp2p-router-abort-handling

cc @achingbrain

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #215, opened 2025-04-21
test: only run gwc test suite once

SgtPooki:
## Title

test: only run gwc test suite once

## Description

No longer run GWC test suite multiple times. Run once, and have one collection of expected passing and failing tests.

Related https://github.com/ipfs/helia-verified-fetch/pull/214

## Notes & open questions

In #214 I had to reduce gwc expected success rate to 45.15% because we started running way more tests, I suspect because signals were finally being respected. Initial investigations on that PR show that the number of tests ran went from ~400 to ~900.

## Change checklist


SgtPooki (commented):
self review

PR #220, opened 2025-04-24
fix: set CORS headers in verified-fetch

SgtPooki:
adds CORS headers to the response.

When adding the CORS headers, many other gateway conformance tests started running, so there are added failing tests, but you can see that there are no tests that were removed from the list of expected passing tests.


packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -211,6 +211,20 @@ export class VerifiedFetch {
...
+    response.headers.set('Access-Control-Allow-Headers', 'Range, X-Requested-With')
+    response.headers.set('Access-Control-Expose-Headers', 'Content-Range, Content-Length, X-Ipfs-Path, X-Stream-Output')
+
+    if (reqFormat !== 'car') {
+      // if we are not doing streaming responses, set the Accept-Ranges header to bytes to enable range requests
+      response.headers.set('Accept-Ranges', 'bytes')
+    } else {
+      // set accept-ranges to none to disable range requests for streaming responses
+      response.headers.set('Accept-Ranges', 'none')
+    }
```
SgtPooki, line 226: this was a quick fix for some "new" failing tests that I looked into briefly.

SgtPooki (commented):
self review

PR #221, opened 2025-04-24
fix: verified-fetch handles OPTIONS method

SgtPooki:
Another fix on top of https://github.com/ipfs/helia-verified-fetch/pull/220


SgtPooki (commented):
self review

Issue #228, opened 2025-05-12
bug: hamt-sharded directory content not returning correct content type


SgtPooki:
only verified in sw-gateway:

related:

https://explore.ipld.io/#/explore/bafybeibc5sgo2plmjkq2tzmhrn54bk3crhnc23zd2msg4ea7a4pxrkgfna/371/
http://ipfs.io/ipfs/bafybeibc5sgo2plmjkq2tzmhrn54bk3crhnc23zd2msg4ea7a4pxrkgfna/371/

```bash
REQUESTS_FILE=/Users/sgtpooki/Downloads/requests-2025-05-07-filtered.json START_INDEX=1000 TEST_COUNT=1000 npm run test:chrome -- -g 'feasibility' --retries 0 --max-failures 999 --workers 2 -g 'bafybeibc5sgo2plmjkq2tzmhrn54bk3crhnc23zd2msg4ea7a4pxrkgfna'
```

<img width="1431" alt="Image" src="https://github.com/user-attachments/assets/dbdf3ce7-82da-48f3-9e21-ae9dc46b7568" />

<img width="1358" alt="Image" src="https://github.com/user-attachments/assets/83459595-69a9-4882-b8de-4f1d7d6b4ca7" />

PR #229, opened 2025-05-12
fix: ending path is not assumed to be filename

SgtPooki:
- **test: reproduce #228**
- **fix: setContentType does not assume path end is filename**
- **fix: pass filename query param to setContentType**

## Title

fix: ending path is not assumed to be filename

## Description

Previously, the `setContentType` function assumed that the ending path of the request was the filename. This is not always the case, as paths may end in some label not reflective of the filename.

Also, we now utilize the `filename` query parameter when setting the content-type, as specified by the [Path Gateway spec](https://specs.ipfs.tech/http-gateways/path-gateway/#filename-request-query-parameter).

Fixes https://github.com/ipfs/helia-verified-fetch/issues/228

## Change checklist


SgtPooki (commented):
self review

PR #244, opened 2025-05-23
fix!: plugins require a unique id

### LINKED ISSUE #243: fix: more clear log prefixes for plugins
add required logPrefix property to BasePlugin so plugins must extend it and we wont have uglified log prefixes for all plugins

SgtPooki:
- **fix: add id for logging and plugin recognition**
- **fix: use id for plugin identification**

## Title

fix!: plugins require a unique id

## Description

Previously when debugging, lognames were not clear because they were based off the class name, which is minified in production builds, resulting in confusing log messages.

Now, we require plugins to have a unique id, and use it for logging and filtering and plugin overrides by consumers.

Fixes #243

## Change checklist


SgtPooki (commented):
self review

Issue #260, opened 2025-07-29
bug: fs.ls seems to query too many blocks


SgtPooki:
We need to investigate how wide dags are handled in @helia/verified-fetch and add some tests to ensure we're not (or can choose not to) unnecessarily load children blocks when doing `fs.ls`.

This issue should probably belong in `@helia/unixfs` but opening here because this is where the issue is observed first.

Potentially related to https://github.com/ipfs/js-ipfs-unixfs/pull/437

---

Based on a [discussion on slack](https://ipshipyard.slack.com/archives/C046HDAHA13/p1753376307057309): 

<details>
<summary>Slack discussion log</summary>

> **Arkadiy - Internet Archive [Thursday at 12:58 PM]**
> ok it looks like the stall on the ls iterator happens due to this:

> **Daniel Norman [Thursday at 12:59 PM]**
> Nice! So we probably just need some better concurrency control

> **Arkadiy - Internet Archive [Thursday at 1:07 PM]**
> I have also identified ~1370 as the magic number that breaks every time; I am guessing the requests are dispatched all at once so that must be hitting some hard ceiling

> **Arkadiy - Internet Archive [Thursday at 1:43 PM]**
> also this is chromium specific firefox actually does the whole batch fine

> **SgtPooki [Thursday at 2:30 PM]**
> IIRC: there is no way to query the browser to determine if the network request queue is full, blocked, nor reaching some limit.Alex investigated this a bit when looking into the WebTransport bug in chrome a while ago

> **Arkadiy - Internet Archive [Friday at 9:25 AM]**
> ok I actually have a deployed test for you

> **SgtPooki [Friday at 9:45 AM]**
> FYI you may want to switch from the default memoryStore to a persistent store: https://github.com/parkan/helia-ia-frontend/blob/3f5382d728e41b1cffc14058def592e6b65eb005/sw/sw.ts#L118-L128

> **Arkadiy - Internet Archive [Friday at 10:07 AM]**
> thanks, this is probably better than my currently hand-managed app-layer cache

> **SgtPooki [Friday at 10:19 AM]**
> I did initially, but I can’t after

> **SgtPooki [Friday at 10:32 AM]**
> yea thats what i got before too.. and if I refresh, it seems to work.one issue a persistent storage should solve for you: re-clicking browse content wouldn’t start over at zero.. it would have all the content it already got

> **Arkadiy - Internet Archive [Friday at 11:33 AM]**
> it already does not start over at zero as far as I can tell

[...]

PR #261, opened 2025-08-05
fix: do not download all blocks when listing dir

SgtPooki:
## Title

fix: do not download all blocks when listing dir

## Description

Fixes https://github.com/ipfs/helia-verified-fetch/issues/260
Related https://github.com/ipfs/helia/pull/836
Related https://github.com/ipfs/js-ipfs-unixfs/pull/439
Related https://github.com/ipfs/js-ipfs-unixfs/pull/440

## Notes & open questions

Two things:

1. There are multiple places where we could potentially use the `extended: false` option.
2. Currently there seems to be a bug or problem with using the ipfs-unix-fs exporter with the `extended: false` option.

### Other places we may want to use `extended: false`

There are multiple places where we use `unix-fs-exporter`.. these were added and commented out in the initial commit in this PR: [`e4d7a15` (#261)](https://github.com/ipfs/helia-verified-fetch/pull/261/commits/e4d7a15dc77f43ce1ce891aeefce708595ba0b48)

#### `./packages/verified-fetch/src/utils/walk-path.ts`

`const entry of exporterWalk(path, blockstore, options)` do we need to call it here?

#### `./packages/verified-fetch/src/plugins/plugin-handle-dag-pb.ts`

places where we call `() => exporter`

### Bug with `extended: false`

Looking at these lines in `plugin-handle-dag-pb`:

https://github.com/ipfs/helia-verified-fetch/blob/cd5a103c122ad447610d4400f5cbf3fda1e8317e/packages/verified-fetch/src/plugins/plugin-handle-dag-pb.ts#L96-L101

When `extended: true`, and added a console.log for `entry`,I get the following output:

```shell
  helia:verified-fetch:dag-pb-plugin:trace found directory at bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm/, looking for index.html +0ms
entry {
  type: 'file',
  name: 'index.html',
  path: 'bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm/index.html',
  cid: CID(bafybeig6e22n6aozgcfdqlh3ysctvsfwrqi7qa6u4qrafb5pzn3keke6ci),
  content: [AsyncGeneratorFunction: yieldFileContent],
  unixfs: UnixFS {
    type: 'file',
    data: undefined,
    blockSizes: [ 2n, 2n ],
    hashType: undefined,
    fanout: undefined,
    mtime: undefined,
    _mode: 420,
    _originalMode: 0
  },
  depth: 1,
  node: {
    Data: Uint8Array(8) [
       8, 2, 24, 4,
      32, 2, 32, 2
    ],
    Links: [ [Object], [Object] ]
  },
  size: 4n
}
  helia:verified-fetch:dag-pb-plugin:trace found root file at bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm/index.html with cid bafybeig6e22n6aozgcfdqlh3ysctvsfwrqi7qa6u4qrafb5pzn3keke6ci
```

When `extended: false`, I get the following:

```shell
  helia:verified-fetch:dag-pb-plugin:trace found directory at bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm/, looking for index.html +0ms
entry {
  cid: CID(bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm),
  name: 'bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm',
  path: 'bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm'
}
  helia:verified-fetch:dag-pb-plugin:trace found root file at bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm/index.html with cid bafybeifgpc3zqsh3ftf6u65rwuid2okrc75f7ojxipzyzlfkepzfxhlavm
```

The exporter is not reporting on the path passed to it.. instead taking the root and seeming to stop. This is tested with:

```shell
DEBUG="-mocha*,*,*:trace" npm run test -- -t node -g 'should abort a request while loading file data'
```

in <helia-verified-fetch repo>/packages/verified-fetch

At the most, this seems like a bug.. at the least, it’s unexpected that passing extended: false would change the CID the response is giving back

**In summary:**

Passing `extended: false` on a path that exists, `<CID>/index.html` in this case, to the exporter returns only the CID of the directory, not of the file at the path I give it.

## Change checklist


packages/verified-fetch/src/utils/dir-index-html.ts
```diff
@@ -165,7 +165,7 @@ export const dirIndexHtml = (dir: UnixFSEntry, items: UnixFSEntry[], { gatewayUR
     },
     listing: items.map((item) => {
       return {
-        size: item.size.toString(),
+        size: item.size?.toString() ?? '?',
```
SgtPooki, line 168: we no longer have the sizes.. we will need to either fetch them with stat or decide on a better default.

packages/verified-fetch/src/utils/walk-path.ts
```diff
@@ -24,7 +24,9 @@ async function walkPath (blockstore: ReadableStorage, path: string, options?: Pa
   const ipfsRoots: CID[] = []
   let terminalElement: UnixFSEntry | undefined
 
-  for await (const entry of exporterWalk(path, blockstore, options)) {
+  for await (const entry of exporterWalk(path, blockstore, {
+    ...options
+  })) {
```
SgtPooki, line 29: undo this change if we're not changing extended's default:


```suggestion
  for await (const entry of exporterWalk(path, blockstore, options)) {
```

SgtPooki (commented):
self review

parkan:
@SgtPooki can I help resolve this with feedback from a user perspective?

SgtPooki:
change in https://github.com/ipfs/js-ipfs-unixfs/pull/440 should fix passing `extended: false` not returning what we expect

SgtPooki:
tested this branch vs main (using `npm run build && DEBUG="*,*:trace" node packages/gateway-conformance/dist/src/demo-server.js`).

@parkan let me know if anything seems off here.. Ideally we would add some configurability to the dir-index-html plugin to support getting the sizes of children when necessary..


### This PR:

#### Server timing (1.78s to load large dir)

image


#### Screenshot
image

### Main:

#### Server timing (times out due to basic-server 10s timeout)

image


#### Screenshot

image

parkan:
beautiful! will test and report but the numbers are already looking good :+1:

PR #365, opened 2026-06-04
fix: add cache-control directives and expires header to responses

achingbrain:
Add the `stale-while-revalidate` and `stale-if-error` `cache-control` header directives to responses that resolve an IPNS name, and also an `expires` header that contains the record EOL.

## Change checklist

### DIFF
--- a/packages/verified-fetch/src/plugins/plugin-handle-ipns-record.ts
+++ b/packages/verified-fetch/src/plugins/plugin-handle-ipns-record.ts
@@ -19,7 +19,7 @@ export class IpnsRecordPlugin extends BasePlugin {
     return accept.some(header => header.contentType.mediaType === MEDIA_TYPE_IPNS_RECORD)
   }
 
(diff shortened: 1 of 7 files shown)


packages/verified-fetch/src/plugins/plugin-handle-ipns-record.ts
```diff
@@ -46,18 +46,31 @@ export class IpnsRecordPlugin extends BasePlugin {
...
+    }
+
+    if (result.record.validityType === 'EOL') {
+      const eol = new Date(result.record.validity)
+      headers.expires = eol.toUTCString()
+
+      const lifetimeRemaining = Math.max(0, Math.round((eol.getTime() - Date.now()) / 1000))
+
+      headers['cache-control'] += `, stale-while-revalidate=${lifetimeRemaining}, stale-if-error=${lifetimeRemaining}`
+    }
```
lidel, line 54: the stale-directive + `Expires` logic is duplicated here and  in  `response-headers.ts:52-60`. A shared helper would keep them from drifting, but also, unlikely we will ever touch this again so not a blocker.

lidel (approved):
LGTM (but check two small fixups I've pushed)

PR #375, opened 2026-08-17
fix: escape invalid characters in x-ipfs-path

### LINKED ISSUE #1185: Setting x-ipfs-path header fails when path contains non-ASCII characters (e.g., 「 」)
### Checklist

- [x] I have searched the [issue tracker](https://github.com/ipfs/service-worker-gateway/issues?q=is%3Aissue) for existing issues.

### Description

it's reproducible at https://bafybeiazzh5mdz3m724mnqkkxspdorekxz5372ck3p3voio26myjnpuf44.ipfs.inbrowser.dev
```js
const path = uint8ArrayToString(uint8ArrayFromString(decodedPath, 'ascii'), 'ascii')
response.headers.set('x-ipfs-path', `/${context.url.protocol === 'ipfs:' ? 'ipfs' : 'ipns'}/${context?.url.hostname}${path === '/' ? '' : path}`)
```

The current implementation attempts to set the x-ipfs-path response header using a path that may contain non-ASCII characters (e.g., CJK punctuation like 「 and 」). The code converts the path via uint8ArrayFromString(decodedPath, 'ascii') and then back to a string, which corrupts non-ASCII data, and ultimately fails when calling response.headers.set() because HTTP header values are restricted to ASCII (or ISO-8859-1 in some implementations, but typically require valid characters).

achingbrain:
Adds `ipfs-uri` header that can correctly encode all path characters.

Wrap setting `x-ipfs-path` and `ipfs-uri` headers in try/catch to prevent bad characters failing the response.

Limits max header value to `8192` bytes.

Fixes: ipfs/service-worker-gateway#1185
Refs: https://github.com/ipfs/specs/issues/547

## Change checklist


packages/verified-fetch/src/utils/ipfs-path-to-header.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Joins an array of strings as an IPFS path and URI encodes individual
+ * components.
+ *
+ * Headers can ony contain extended ASCII but IPFS paths can be unicode.
+ */
+export function ipfsPathToHeaderValue (protocol: string, hostname: string, path: string): string {
```
lidel: some drive-by nits @achingbrain, but I think encoding only characters above 255 leaves a few problems and it may be more headache than just doing the simple thing:

1. `split('')` breaks emoji into surrogate halves, and `encodeURIComponent` throws `URIError` on a lone surrogate. This happens outside the try/catch, so a fetch of `/😀.txt` fails instead of returning a response.
2. The output cannot be decoded back. Characters like `é` stay literal, higher ones become percent-escapes, and a literal `%` is never re-encoded, so `/100%25off.txt` comes out as `/100%off.txt`. Readers that percent-decode this header (service-worker-gateway does) will get wrong values.
3. A path with `%0A` decodes to a real newline, `Headers.set` still throws, and the header is silently dropped. Also, `.trim()` changes filenames that start or end with a space.

I think there is a simpler fix that avoids all of this: skip the decode step and use `context.url.pathname` as-is. WHATWG URLs already percent-encode it (`/「1」.txt` serializes as `/%E3%80%8C1%E3%80%8D.txt`), so the value is lossless, always valid in a header, and needs no try/catch. It also matches where ipfs/specs#547 is heading: value is fully percent-encoded, readers percent-decode. If we go that way, the one visible change is that spaces become `%20`, but that is the future-proof spec direction anyway I guess?

achingbrain: All good points. I have updated this PR to use `ipfs-uri` over `x-ipfs-path` as per the discussion in https://github.com/ipfs/specs/issues/547.

PR #376, opened 2026-08-21
feat: canonical Ipfs-Uri header (IPIP-548)

lidel:
## Problem

Since #375 the `ipfs-uri` header echoes the request's `url.pathname`, so its value depends on how the caller spelled the path: raw `+` or `=` and lowercase hex pass through, and two clients requesting the same file can see different values. `x-ipfs-path` is still sent with non-ASCII names squashed to ASCII bytes, which corrupts them. ipfs/specs#548 now pins one canonical serialization.

## Fix

- build `ipfs-uri` per ipfs/specs#548: canonical authority (base32 CIDv1, base36 IPNS name, A-label DNSLink), each path segment decoded once and strictly re-encoded, no query or fragment
- keep `x-ipfs-path` for legacy clients, but omit it when the path cannot appear in an HTTP field value instead of mangling it
- expose `Ipfs-Uri` for CORS

Similar behavior as boxo (ipfs/boxo#1209); both are covered by the shared fixtures in ipfs/gateway-conformance#301.

Feel free to simplify if being 100% spec compliant and covering all edge cases is not as important as smaller code surface.


packages/verified-fetch/test/ipfs-uri.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+    it('should keep a trailing slash after a directory path', () => {
+      expect(toIpfsUri(new URL(`ipfs://${cid}/dir/`))).to.equal(`ipfs://${cid}/dir/`)
+    })
+
+    it('should drop interior empty segments', () => {
+      expect(toIpfsUri(new URL(`ipfs://${cid}/a//b//`))).to.equal(`ipfs://${cid}/a/b/`)
+    })
+
+    it('should treat an encoded slash as a path separator', () => {
+      expect(toIpfsUri(new URL(`ipfs://${cid}/a%2Fb`))).to.equal(`ipfs://${cid}/a/b`)
```
achingbrain: If we have to decode each segment like this, how do you represent a filename with a `/` character in it?

lidel: I think the honest answer is: You can't, same as POSIX: `/` is the separator. A dag-pb link named `a/b` is legal bytes, but the UnixFS spec's [Paths section](https://specs.ipfs.tech/unixfs/#unixfs-paths) already forbids it in paths ("Components MUST NOT contain `/`"), so such an entry is reachable only by its own CID. Go gateways also decode `%2F` before the handler ever sees it, so kubo has always read it as a separator; reserving `%2F` for literal slashes would make the same URL address different files on old vs new gateways.

lidel: IPIP-548 seems like a good opportunity to make that explicit across HTTP gateways, UnixFS restricted names, and the URI serialization +  added with fixture to it: https://github.com/ipfs/specs/pull/548/commits/01ababbcc44e5d7b085c5113be3db5a999e92350
Does that look sensible? @achingbrain

achingbrain: Ah yes, fair enough - I was looking for that in the spec but I missed it.  Looks good.

packages/verified-fetch/test/ipfs-uri.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+
+      expect(toIpfsUri(new URL(`ipns://${cid}`))).to.equal(undefined)
+    })
+
+    it('should strip a single trailing dot from a DNSLink name', () => {
+      expect(toIpfsUri(new URL('dnslink://dnslink.example.net./hello'))).to.equal('ipns://dnslink.example.net/hello')
+    })
+
+    it('should omit the header when a DNSLink name has no dot', () => {
+      expect(toIpfsUri(new URL('dnslink://localhost/hello'))).to.equal(undefined)
```
achingbrain: This is a valid hostname, though?  Putting `localhost` aside for a second, you could be on a private network with internal DNS resolving DNSLink entries for local servers?

lidel: Good question. A gateway can still resolve DNSLink for such hosts; the dot rule only controls what goes into `Ipfs-Uri`. A name without a TLD means different things on different networks, so it cannot be a stable `ipns://` authority, and [ipns-uri](https://specs.ipfs.tech/ipns-uri/) defines such names as invalid (TL;DR: a DNSLink name needs a dot).

So no TLD means no header. For local DNSLink, use `.local` ([RFC 6762](https://www.rfc-editor.org/rfc/rfc6762)).

I switched tests to `example.local` (positive) and `examplemissingtld` (negative) in ba94171; same pair in boxo, gateway-conformance, and ipfs/specs@744a94b.

achingbrain (approved):

PR #379, opened 2026-09-03
fix: fall back to the filename extension for common media types

Chara-Freedom:
## Problem

`contentTypeParser` sniffs with `file-type` first, which only looks for magic bytes at offset 0. A file that opens with padding is not recognised, so it falls through to the filename-extension switch — and that switch covers css, html, js, json, txt, woff2, svg, csv, doc, xls, ppt and msi only. Everything else becomes `application/octet-stream`, which browsers download instead of playing or displaying inline.

This was reported downstream in ipfs/service-worker-gateway#1197: two near-identical MP3s in the same directory, one plays and the other downloads. The one that downloads has 417 zero bytes before its first MPEG frame sync, so `fileTypeFromBuffer` returns `undefined` and `.mp3` is not in the switch.

The reference gateway does not have this problem — Kubo serves both files as `audio/mpeg`, because Go's `http.ServeContent` maps the extension first and only sniffs when the extension is unknown:

```console
$ curl -sI https://ipfs.io/ipfs/bafybeibce3vhbe6fn5jyiiyc7pzg2vxuqquxeigdqiuqxrt67f6yy3cwse/police.mp3 | grep -i content-type
Content-Type: audio/mpeg
```

## Change

- Adds the common audio (`mp3`, `m4a`, `aac`, `flac`, `wav`, `ogg`/`oga`/`opus`, `weba`, `mid`/`midi`), video (`mp4`, `m4v`, `webm`, `mkv`, `mov`, `avi`, `ogv`) and document (`pdf`, `ico`, `md`, `vtt`, `xml`) extensions to the switch. Several of these are formats `file-type` normally recognises — they are listed because sniffing is exactly what fails when the magic bytes are not at offset 0.
- Matches the extension case-insensitively, so `POLICE.MP3` resolves the same as `police.mp3`.

Sniffing still takes precedence, so a PNG named `.mp3` is unaffected — there is a test for that.

## Tests

Four cases added to `test/content-type-parser.spec.ts`: a padded MP3 (with an assertion that `fileTypeFromBuffer` really does give up on those bytes), the case-insensitive match, sniffed bytes winning over a misleading extension, and `.md`.

```
17 passing  (aegir test -t node --grep content-type-parser)
```

`aegir lint` and `aegir doc-check` are clean, and the package builds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/verified-fetch/src/utils/content-type-parser.ts
+++ b/packages/verified-fetch/src/utils/content-type-parser.ts
@@ -60,8 +60,15 @@ export async function contentTypeParser (bytes: Uint8Array, fileName?: string): 
     return defaultMimeType
   }
 
-  // no need to include file-types listed at https://github.com/sindresorhus/file-type#supported-file-types
-  switch (fileName.split('.').pop()) {
+  // formats file-type cannot sniff at all, plus the media containers it only
+  // recognises when their magic bytes start at offset 0 - a file with leading
+  // padding (an MP3 whose first frame sync sits behind a run of silence, a PDF
+  // that does not begin with its header) reaches this switch as well, and
+  // `application/octet-stream` makes the browser download it instead of
+  // playing or displaying it inline
+  // @see https://github.com/sindresorhus/file-type#supported-file-types
+  // @see https://github.com/ipfs/service-worker-gateway/issues/1197
+  switch (fileName.split('.').pop()?.toLowerCase()) {
     case 'css':
       return 'text/css'
     case 'html':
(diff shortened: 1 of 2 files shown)


Chara-Freedom:
The `check` failure was mine — `msvideo` and a British spelling. Fixed in bbbce0d.

The remaining test failure is not from this PR. Every job that actually runs the suite reports the same 492 passing / 1 failing:

```
@helia/verified-fetch > 404 paths > returns a 404 when requesting CID path from dag-pb IPFS URL:
  AssertionError: expected 502 to equal 404
```

That `describe` stubs `contentTypeParser` with a Sinon stub, so none of the code this PR touches is executed there.

It is a regression in `ipfs-unixfs-exporter` 16.2.0: walking a path that continues below a dag-pb file node now throws `NotUnixFSError` / `ERR_NOT_UNIXFS`, where 16.1.0 threw `NotFoundError` / `ERR_NOT_FOUND`. `errorToResponse` maps the former to a 502 Bad Gateway and the latter to a 404, hence the assertion. Reproduction, version bisect and the throw site are in ipfs/js-ipfs-unixfs#492.

There is no committed lockfile and the dependency is `^16.0.2`, so `main` is affected too — the last green run on `main` is from 2026-08-27, and 16.2.0 was published 2026-09-10.

One more thing worth knowing when reading the check list: `test-deno`, `test-bun`, `test-webkit`, `test-webkit-webworker` and `test-electron-renderer` are green vacuously — those scripts are not defined in the root `package.json`, so `npm run --if-present` is a no-op. The real coverage is the six jobs that failed.

Chara-Freedom:
@achingbrain

achingbrain:
Thanks!  You should also open an issue on `file-type` so it can detect these files more reliably.

Chara-Freedom:
> Thanks! You should also open an issue on `file-type` so it can detect these files more reliably.

Done, and waiting for the service-worker release.

Chara-Freedom:
> Thanks! You should also open an issue on `file-type` so it can detect these files more reliably.

They replied with a recommendation to use mpegOffsetTolerance https://github.com/sindresorhus/file-type/issues/798#issuecomment-5657106818, apparently this is a popular problem so they already have a built-in function.

